PR #10, opened 2022-11-15
(chore) Fix coverage script

denniskigen:
Removes an unnecessary `--` from the `--coverage` flag passed to `yarn test` in the coverage script.


denniskigen:
Looks like this was only required following the yarn 3 upgrade, so I've fixed it within that PR.

PR #11, opened 2022-11-15
(chore) Configure ESLint rules for carbon and lodash imports

denniskigen:
Sets up ESLint rules that govern the imports of carbon and lodash dependencies. Specifically, these rules will trigger failures in the lint check if a developer attempts to import carbon components using the old (v10) approach. The same applies to imports from the `lodash` or `lodash-es` libraries.


gracepotma:
Nice, neat idea

pirupius (approved):
🚀

PR #23, opened 2022-12-01
(chore) Monitor bundle sizes using compressed-size-action

denniskigen:
Sets up `compressed-size-action` to monitor bundle size changes across pull requests.


denniskigen:
Trivial; merging.

PR #25, opened 2022-12-01
(feat) Use props instead of context for state management

denniskigen:
This PR migrates the state management approach used in this repo from context to props. Whilst it's tempting to reach for context to manage state, I think this application's state management needs [do not merit](https://beta.reactjs.org/learn/passing-data-deeply-with-context#before-you-use-context) using context. 

The significant benefit of using props is that they make data flow explicit, which can help with maintainability. A common concern when passing props is the dreaded [prop-drilling](https://kentcdodds.com/blog/prop-drilling) problem. I don't think that's a problem here because all the components that receive the props use them (rather than only passing them further down).

### Further reading

- https://kentcdodds.com/blog/application-state-management-with-react
- https://kentcdodds.com/blog/how-to-use-react-context-effectively

Note, however, that this is a strong view that is weakly held. I'm very open to alternative arguments.

### DIFF
--- a/src/components/form-editor/element-editor/element-editor.tsx
+++ b/src/components/form-editor/element-editor/element-editor.tsx
@@ -1,16 +1,22 @@
-import React, { useContext, useEffect, useState } from "react";
-import { Accordion, Button, Column, Row, TextInput } from "@carbon/react";
+import React, { useEffect, useState } from "react";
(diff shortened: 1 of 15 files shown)


pirupius (approved):
Thanks for the links in the description. Very insightful 💯

PR #29, opened 2022-12-01
(fix) Amend patientUuid config property

denniskigen:
Amends the `patientUuid` config property as follows:

- Renames the property from `patientUuidConfig` to just `patientUuid`.
- Changes the default UUID to that of an existing test patient on the dev3 server.
- Updates the property description to accurately represent what it actually does.

### DIFF
--- a/src/config-schema.ts
+++ b/src/config-schema.ts
@@ -37,9 +37,10 @@ export const configSchema = {
       "problem",
     ],
   },
-  patientUuidConfig: {
+  patientUuid: {
     _type: "String",
-    _default: "b280078a-c0ce-443b-9997-3c66c63ec2f8",
-    _description: "Patient UUID used to render form preview",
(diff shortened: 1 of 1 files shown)


src/config-schema.ts
```diff
@@ -37,9 +37,10 @@ export const configSchema = {
       "problem",
     ],
   },
-  patientUuidConfig: {
+  patientUuid: {
     _type: "String",
-    _default: "b280078a-c0ce-443b-9997-3c66c63ec2f8",
-    _description: "Patient UUID used to render form preview",
+    _default: "88f1032f-adae-4ef2-9025-2c40b71dd897",
```
pirupius, line 42: Is the patient with this UUID always there even with the various server outages and restarts?

denniskigen: This one has proven to be a bit more resilient than most.

pirupius (approved):

PR #33, opened 2022-12-06
(refactor) Code improvements

denniskigen:
This PR makes wide-ranging improvements to code, including:

- Renaming some variables so their names better describe what they do.
- Caching some function definitions using `useCallback` and `useMemo` hooks.
- Renaming a CSS class.


src/components/dashboard/dashboard.component.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { useEffect, useMemo, useState } from "react";
+import React, { useMemo, useState } from "react";
```
pirupius, line 1: The movement against `useEffect` has started ✊🏾

pirupius (approved):

PR #35, opened 2022-12-11
(feat) Editor mode

denniskigen:
This PR augments the interactive schema builder with the ability to edit:

- the schema name
- pages
- sections
- questions

## Video

https://user-images.githubusercontent.com/8509731/206930214-f853cbef-d787-4e51-89bb-0a4af13c2e0d.mp4

### DIFF
--- a/src/components/interactive-builder/interactive-builder.component.tsx
+++ b/src/components/interactive-builder/interactive-builder.component.tsx
@@ -1,6 +1,6 @@
-import React, { useEffect, useState } from "react";
+import React, { useState } from "react";
 import { useTranslation } from "react-i18next";
-import { Button, InlineLoading } from "@carbon/react";
+import { Accordion, AccordionItem, Button, InlineLoading } from "@carbon/react";
 import { Add, Edit } from "@carbon/react/icons";
 import { useParams } from "react-router-dom";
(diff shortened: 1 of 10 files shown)


src/components/interactive-builder/interactive-builder.component.tsx
```diff
@@ -185,22 +299,43 @@ const InteractiveBuilder: React.FC<InteractiveBuilderProps> = ({
                 ) : null}
                 {page?.sections?.length ? (
                   page.sections?.map((section, sectionIndex) => (
-                    <Accordion className={styles.accordion}>
-                      <AccordionItem title={section.label}>
```
pirupius, line 189: @denniskigen Does the css in these div replicate the behaviour of the accordion? Long forms benefited from expanding and collapsing sections.

denniskigen: It doesn't, unfortunately. I haven't thought of a better UI for editing section labels when using Accordions. One way would be to have the section's label as an editable field (alongside its related questions) within an Accordion item.

denniskigen: Accordions are back

pirupius (approved):

PR #43, opened 2022-12-23
Fixed issue with yarn pack ignoring dist

FlorianRappl:
Removed the unnecessary (invalid) upload of artifacts from the workflow and added the `files` explicitly that `yarn pack` picks them up. Otherwise, the necessary chunks are missing and the MF will not load.


denniskigen:
Thank you, @FlorianRappl!

PR #47, opened 2023-01-02
(feat) Resolve concept names from UUIDs when editing questions

denniskigen:
Enables concept name resolution when editing a question. This enables us to grab the concept UUID of the question being edited from the schema, look up its concept name and display that name in the `Select backing concept` field of the form. This means we can now display the name of the concept rather than it's UUID (which isn't very useful to the user).  

> Concept name resolution when editing questions

https://user-images.githubusercontent.com/8509731/210528081-92271f6c-d283-472f-b75c-1f76bee0bb21.mp4


pirupius (approved):
🚀

PR #52, opened 2023-01-07
(feat) Display indent guides in schema editor

denniskigen:
Restores indent guides in the schema editor, which should help with bracket pair matching and visually parsing code blocks.

Screenshot 2023-01-07 at 14 14 15


pirupius (approved):
Not sure how i missed this 🤔

PR #54, opened 2023-01-11
(feat) Add the ability to duplicate questions

denniskigen:
## Summary

This PR adds the ability to duplicate questions using the interactive builder. The duplicated question gets a placeholder question ID which ought to be changed by the user. This is meant to help the user scaffold out a form quickly.

Additionally, this PR also ups the `enterDelayMs` property value of several icon buttons to 200ms, ensuring that icon tooltips don't overlap. 

## Video

https://user-images.githubusercontent.com/8509731/211738657-07bf851d-ea88-43fb-bd4b-4215f0e65fb3.mp4

### DIFF
--- a/src/components/interactive-builder/add-question-modal.component.tsx
+++ b/src/components/interactive-builder/add-question-modal.component.tsx
@@ -173,69 +173,6 @@ const AddQuestionModal: React.FC<AddQuestionModalProps> = ({
     }
   };
 
-  const updateQuestion = (questionIndex) => {
-    try {
-      const mappedAnswers = selectedAnswers?.map((answer) => ({
-        concept: answer.id,
-        label: answer.text,
-      }));
-
-      const data = {
-        label: questionLabel ? questionLabel : questionToEdit.label,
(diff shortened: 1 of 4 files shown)


src/components/interactive-builder/add-question-modal.component.tsx
```diff
@@ -173,69 +173,6 @@ const AddQuestionModal: React.FC<AddQuestionModalProps> = ({
     }
   };
 
-  const updateQuestion = (questionIndex) => {
-    try {
-      const mappedAnswers = selectedAnswers?.map((answer) => ({
-        concept: answer.id,
-        label: answer.text,
-      }));
```
pirupius, line 181: Has this been moved to somewhere else? or are u going for the duplicate question approach?

denniskigen: This logic got moved to the `EditQuestionModal` component - so what's left over here is the cruft that ought to have been deleted.

pirupius (approved):

PR #57, opened 2023-01-12
(chore) Add release script

denniskigen:
Adds a release script that can be used to cut releases of this project.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
     "start": "openmrs develop",
     "serve": "webpack serve --mode=development",
     "build": "webpack --mode production",
+    "release": "lerna version --no-git-tag-version",
     "analyze": "webpack --mode=production --env.analyze=true",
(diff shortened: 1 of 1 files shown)


denniskigen:
@ibacher, what do you reckon would be a good version number to start releasing this project at?

ibacher:
Honestly, I think 1.0.0 is fine... Releasing things starting at 3.0.0 was a mistake...

package.json
```diff
@@ -10,6 +10,7 @@
     "start": "openmrs develop",
     "serve": "webpack serve --mode=development",
     "build": "webpack --mode production",
+    "release": "lerna version --no-git-tag-version",
```
ibacher, line 13: This isn't a lerna repo is it? Is this going to work?

denniskigen: Oops, no.

PR #59, opened 2023-01-12
(chore) Set auth config before running release command

denniskigen:
One more fix, @ibacher, this time to set the authToken before executing the `release` job.


ibacher (approved):

ibacher:
Ah yes... I forgot that that was a necessary step...

PR #61, opened 2023-01-19
O3-1792: Fix issues with retrieving form resource where there exists both html and JSON form resource

donaldkibet:
### What does this PR do.

1. Fix error while trying to retrieve form resource where we have both `JSON` and `HTML` form resource. For scenarios where users don't what to create new forms but use existing `formUuid` and add the JSON resource to enable use of htmlforms 

### Screenshoot 
image

As seen above the form has both `hfexmlPath` and `AmpthJsonSchema` resources for `Alcohol  Abuse Screening form`

### DIFF
--- a/src/hooks/useClobdata.ts
+++ b/src/hooks/useClobdata.ts
@@ -3,9 +3,11 @@ import { openmrsFetch } from "@openmrs/esm-framework";
 import { Form, Schema } from "../types";
 
 export const useClobdata = (form?: Form) => {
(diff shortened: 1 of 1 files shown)


denniskigen (approved):
LGTM, though I can't test this in the dev3 env. Mind posting a screenshot of what the list looked like prior to the fix?

ibacher (approved):
The logic here is correct (especially when we start adding additional form resources). I'm not sure having an HFE form and another form share a UUID is a great idea unless they are very carefully kept in sync.

PR #67, opened 2023-03-21
(fix) Remove explicit use of axios

ibacher:
Fixes two things:

1. Auth failure should redirect to login
2. Backend may not be installed on `/openmrs/`


denniskigen (approved):
LGTM. Thanks!

PR #68, opened 2023-03-24
(fix) Append `multipart/form-data` content type to request headers

denniskigen:
Potentially fixes a bug where form schemas would fail to upload schema blobs to the `clobdata` endpoint.


donaldkibet (approved):
LGTM Thanks @denniskigen

Issue #75, opened 2023-04-06
Unable to edit schema


ckote:
The engine is currently giving an error when you try to edit and save a form

![image](https://user-images.githubusercontent.com/110154947/230301791-ed4cb2f6-27a0-4603-bbf9-8eca20c691c0.png)

PR #76, opened 2023-04-07
(fix) Fix errors when attempting to save a form

denniskigen:
Fixes an issue where updating existing forms via the `Save as Existing form` feature didn't work reliably. Closes https://github.com/openmrs/openmrs-esm-form-builder/issues/75.

### DIFF
--- a/src/components/modals/save-form.component.tsx
+++ b/src/components/modals/save-form.component.tsx
@@ -23,7 +23,7 @@ import {
   updateName,
   updateVersion,
   updateDescription,
-  getResourceUUID,
+  getResourceUuid,
   deleteClobdata,
   deleteResource,
   updateEncounterType,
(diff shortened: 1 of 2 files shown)


src/components/modals/save-form.component.tsx
```diff
@@ -120,12 +121,27 @@ const SaveForm: React.FC<SaveFormModalProps> = ({ form, schema }) => {
...
       try {
-        if (form?.resources.length != 0) {
-          deleteClobdata(form?.resources[0].valueReference);
-          deleteResource(form?.uuid, form?.resources[0].uuid);
+        if (form?.resources?.length != 0) {
+          const valueReference = form?.resources[0].valueReference ?? "";
+
+          deleteClobdata(valueReference)
+            .catch((error) =>
+              console.error("Unable to delete clobdata: ", error)
```
donaldkibet, line 129: I think it would be best to alert the user that an error has occurred, by either showing the `Toast` or `Notification`

donaldkibet (approved):
LGTM Thanks @denniskigen

src/components/modals/save-form.component.tsx
denniskigen: Sure

PR #84, opened 2023-04-13
(fix) Update test patient UUID

denniskigen:
Necessitated by the recent dev3 reset.

### DIFF
--- a/src/config-schema.ts
+++ b/src/config-schema.ts
@@ -40,7 +40,7 @@ export const configSchema = {
   },
   patientUuid: {
     _type: "String",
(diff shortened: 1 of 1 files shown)


ibacher (approved):

ibacher:
At some point, we should refactor this so that the sample patient used for these purposes is part of the form builder itself or the user can select a patient of interest because doing this via config like this is very fragile and there is no way this will work repeatable across different installations.

denniskigen:
Issue filed https://github.com/openmrs/openmrs-form-engine-lib/issues/20.

PR #85, opened 2023-04-13
(feat) Show a warning about saving schemas

denniskigen:
Renders a warning message in the form builder dashboard about the possibility of users losing their work and suggests saving schemas locally or to the distro repo.

## Screenshot

Screenshot 2023-04-13 at 11 57 02 PM


denniskigen:
Don't know if its best to show this in the schema editor view rather than here in the dashboard.

src/components/dashboard/dashboard.scss
```diff
@@ -110,3 +110,6 @@
   align-items: center;
 }
 
+.warningMessage {
+  margin: 1.5rem 0;
+}
```
ibacher, line 115:
```suggestion
}

```

ibacher (approved):
I love it! One tweak, can we put this behind some configuration that we can enable on dev3 etc., just so this message doesn't show up in prod?

ibacher (approved):

Issue #105, opened 2023-04-29
Forms get Stuck in Loading State with "Update Existing Version"


gracepotma:
Steps:
1. Go into a form (e.g. Laboratory Tests 2.0)
2. Make a change to that form
3. Decide you want to save it as the existing version --> click Save Form, then Update Existing Version
4. See a success toast saying the Form has Saved, but also see that the UI now shows Loading circles across the app
5. Try going back to the Form Builder home page, then going back to this form
6. See that your changes did indeed save, but the Form remains stuck in a Loading State. So now you cannot preview or edit it any further.  

Video: 

https://user-images.githubusercontent.com/67400059/235302127-c9e30a48-5106-42d0-a6d4-55a282271e42.mov

Issue #104, opened 2023-04-29
Version Number in JSON Schema should be updated when Version Number is changed


gracepotma:
Replication Steps: 
1. Go to an existing form 
2. Make some changes
3. Save your changes as a "new version", with a new version number
4. Notice that the version number in the JSON schema did not get updated --> This is especially relevant for forms you are about to save in your config files, because you can easily end up with many different form version numbers, but they are all the same version in the JSON! 

Example in dev3 right now: Laboratory Tests version 2.1 was successfully saved as version 2.1, but in the JSON, you can see it is still written as "version 2". 
<img width="713" alt="image" src="https://user-images.githubusercontent.com/67400059/235301771-7a655ed7-e3a7-4bc1-bfa0-7f39b812c8b4.png">
<img width="713" alt="image" src="https://user-images.githubusercontent.com/67400059/235301780-06b20218-ff12-420f-bbfe-9811544c6cd5.png">

Expected: Whatever version number the user enters in the Save Details Modal should automatically update the "version" tag in the JSON file, when the user clicks "Save" on the modal. 

Let me know if this is not clear :)

Issue #91, opened 2023-04-19
There should be a better way to deal with missing or corrupt schemas


denniskigen:
The `import schema` feature in the forms dashboard needs to be reworked. As currently configured, it doesn't help users recover their schemas. It should be replaced with a more robust alternative.

PR #107, opened 2023-04-30
(fix) Various fixes and improvements

denniskigen:
## Requirements

- [ ] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-dev.docs.openmrs.org/#/getting_started/contributing?id=your-pr-title-should-indicate-the-type-of-change-it-is) label. See existing PR titles for inspiration.
- [ ] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide) and [design documentation](https://zeroheight.com/23a080e38/p/880723-introduction).

## Summary

This PR addresses the issues raised in #91, #104, and #105. Specifically, it:

- Tweaks the draft schema feature so that loading draft schemas from LocalStorage is only possible when an existing form is launched in the form editor in edit mode, and that form does not have an associated schema. This could be because the schema doesn't exist or if it got corrupted.
- Alters the logic behind how loading states get shown in the form editor = this was likely why the form editor got stuck in a loading state when loading a form (especially after saving a modified schema to the server).
- Ensures that changes to the form's `name`, `version`, `description`, and `encounterType` properties get propagated to the server.
- Improvements to the save workflow - including ensuring the success toast doesn't get shown until saving completes successfully. Additionally, we ensure that the revalidation request for the form only gets fired after the schema gets saved successfully.


src/components/schema-editor/schema-editor.component.tsx
```diff
@@ -40,8 +40,12 @@ const SchemaEditor: React.FC<SchemaEditorProps> = ({
 
   const inputDummySchema = useCallback(() => {
     const dummySchema: OHRIFormSchema = {
-      encounterType: "",
+      encounterType: "Test",
```
samuelmale, line 43: Something minor that crossed my mind: Assuming we manage to implement a schema validator in the future and we require this attribute, we may be force to do something like:

```js

if (isNotEmpty(rootFormNode.encounterType) && rootFormNode.encounterType !== 'Test') {
   // pass
}

```

We need to be more careful with some of these defaults. The same case applies for the `formUuid` attribute.

src/components/schema-editor/schema-editor.component.tsx
```diff
@@ -40,8 +40,12 @@ const SchemaEditor: React.FC<SchemaEditorProps> = ({
...
       name: "Sample Form",
+      processor: "EncounterFormProcessor",
+      referencedForms: [],
+      uuid: "f7a2a8f3-2ba9-49fb-a0d8-aaab0edbc1f0",
```
samuelmale, line 47: Was this change needed?

samuelmale (approved):
Looks great overall, left a few minor comments.

src/components/schema-editor/schema-editor.component.tsx
denniskigen: In hindsight, no. Thanks for catching this.

PR #125, opened 2023-05-18
(fix) O3-2124: Unable to save a form after initial save

denniskigen:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-dev.docs.openmrs.org/#/getting_started/contributing?id=your-pr-title-should-indicate-the-type-of-change-it-is) label. See existing PR titles for inspiration.
- [x] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide) and [design documentation](https://zeroheight.com/23a080e38/p/880723-introduction).

## Summary

This PR fixes an issue where edits made to a form schema via the schema editor would disappear after the `render changes` button gets clicked. This was likely due to the changes to the `useEffect` login in the `FormEditor` component introduced in https://github.com/openmrs/openmrs-esm-form-builder/pull/107/files.

This PR also fixes how the Save Modal obtains its values from the form and schema objects. Following these changes, the form save workflow should now more reliably pick the `name`, `version`, `uuid`, `encounterType`,  and `description` properties of a form from the provided props. 

It also adds a call to  `mutate()` which gets invoked after saving a new form so that the form gets revalidated immediately. The benefit of this is that the schema editor can pick up any changes to the form so that when we edit it again, we have access to the most up-to-date version of the form.

## Related issue

https://issues.openmrs.org/browse/O3-2124

### DIFF
--- a/src/components/form-editor/form-editor.component.tsx
+++ b/src/components/form-editor/form-editor.component.tsx
@@ -54,13 +54,22 @@ const FormEditor: React.FC = () => {
     formUuid && (isLoadingClobdata || isLoadingForm);
 
   useEffect(() => {
-    if (formUuid && clobdata && Object.keys(clobdata).length > 0) {
-      setSchema(clobdata);
-      localStorage.setItem("formJSON", JSON.stringify(clobdata));
-    } else if (formUuid && !isLoadingFormOrSchema && !schema) {
-      setShowLoadDraftSchemaModal(true);
+    if (formUuid) {
+      if (
+        !isLoadingFormOrSchema &&
+        clobdata &&
+        Object.keys(clobdata).length > 0
+      ) {
+        setSchema(clobdata);
+        localStorage.setItem("formJSON", JSON.stringify(clobdata));
+      } else if (
+        !isLoadingFormOrSchema &&
+        (!clobdata || Object.keys(clobdata).length === 0)
+      ) {
+        setShowLoadDraftSchemaModal(true);
+      }
     }
-  }, [clobdata, formUuid, isLoadingFormOrSchema, schema]);
+  }, [clobdata, formUuid, isLoadingFormOrSchema]);
 
   const updateSchema = useCallback((updatedSchema) => {
     setSchema(updatedSchema);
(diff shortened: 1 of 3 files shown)


denniskigen:
There's definitely something wonky going on with the schema editor. I'll have a look tomorrow.

src/components/modals/save-form.component.tsx
```diff
@@ -82,15 +85,22 @@ const SaveForm: React.FC<SaveFormModalProps> = ({ form, schema }) => {
     }
   }, []);
 
-  const handleSubmit = async (event) => {
+  const handleSubmit = async (event: SyntheticEvent) => {
```
ibacher, line 88: Couldn't this be specified as `SyntheticEvent<{ name: { value: string };}>` etc?

src/components/form-editor/form-editor.component.tsx
```diff
@@ -54,13 +54,22 @@ const FormEditor: React.FC = () => {
...
-    } else if (formUuid && !isLoadingFormOrSchema && !schema) {
-      setShowLoadDraftSchemaModal(true);
+    if (formUuid) {
+      if (
+        !isLoadingFormOrSchema &&
+        clobdata &&
+        Object.keys(clobdata).length > 0
+      ) {
+        setSchema(clobdata);
+        localStorage.setItem("formJSON", JSON.stringify(clobdata));
```
ibacher, line 64: This reminds me: we should probably have some naming conventions for localStorage items, especially given this is a MF project, e.g., `@openmrs/esm-form-builder:formJson` or something. (Don't need to do this now, just flagging the discussion point)

src/types.ts
```diff
@@ -67,6 +67,7 @@ export interface Schema {
   uuid: string;
   encounterType: string;
   referencedForms: [];
+  version?: string;
```
ibacher, line 70: Neither here nor there, but for forms it might make sense to take the version from the Form object rather than the schema (it's required for the Form object; it's optional in the schema).

ibacher (approved):
Thanks @denniskigen! LGTM, with one small typing adjustment.

src/types.ts
denniskigen: When the user creates a new form (say by using the `input dummy schema` feature) and explicitly types in a version number in the schema editor, it wont get picked up by the form saving workflow if we don't provide `schema?.version || form?.version` as the possible values for the `Version` field. This isn't a problem when editing an existing form because the `version` gets picked from the form object.

ibacher: Yeah, that makes sense!

PR #172, opened 2023-08-01
(feat) Form validator built into interactive form builder

arodidev:
## Summary
A validation tool that validates whether forms have UUIDS, checks for concepts used the in forms in the backend using the concept UUID or concept mappings provided in the form, and validates whether the datatype on the concepts align with the control type assigned in the form.

## Screenshots
Screenshot 2023-08-14 at 13 47 12


src/components/interactive-builder/interactive-builder.component.tsx
```diff
@@ -204,15 +209,134 @@ const InteractiveBuilder: React.FC<InteractiveBuilderProps> = ({
...
+              questionObject.questions.forEach((question) => {
+                handleQuestionValidation(question);
+              });
+            }
+            handleQuestionValidation(questionObject);
+          }
+        }
+      }
+    } else {
+      console.log("Empty form!");
```
samuelmale, line 238: According to our code formatting conventions, it's not a good practice for production-ready code to have console log statements. That being said, our code linter has been configured to reject those by default. Perhaps you intended to use a `console.warn()` or `console.error()` instead?

src/components/interactive-builder/interactive-builder.component.tsx
```diff
@@ -204,15 +209,134 @@ const InteractiveBuilder: React.FC<InteractiveBuilderProps> = ({
...
+            } else {
+              resolver(
+                conceptObject,
+                `Concept "${conceptObject.questionOptions.concept}" not found`
+              );
+            }
+          })
+          .catch((error) => {
+            console.log(error.message);
+          })
```
samuelmale, line 274: Wrap this logic with an API function and harbor it within the API file. This ensures separation of concerns.

samuelmale (changes requested):
Nice work Jammie! My general comment is: let's separate concerns here. Let's have the validation logic to exist independent of the interactive builder component in-order to keep this component simple.

src/components/interactive-builder/interactive-builder.component.tsx
ebambo, line 238: @arodidev since you are now showing the errors / warnings on the screen why don't you also show this as an wringing on the screen. Am saying this because one can just think the button is not working if they don't look at the console logs.

arodidev: This is well noted, perhaps I'll just display it somewhere on the UI as suggested by @ebambo

src/components/interactive-builder/interactive-builder.component.tsx
arodidev: Noted.

src/form-validator.ts
ebambo, line 1: @arodidev thanks for working on this. The separation of concerns is now done but am not sure if this validator belongs to the `form-builder` or the `form-engine-lib` . Am saying this because the decoupling you did is not just for readability purposes but also for reusability, remember I mentioned to you that would be nice to have this on the `encounter-list.component` to hide or show a warning / error if a form is not valid? @samuelmale please share your point of view please. Am saying `form-engine-lib` because I believe that way it would be easy to call this code here and everywhere else as long as the engine-lib is a dependency on the given context.

ebambo: My comment should not prevent this code to be merged unless moving the code to the definitive place doesn't take long. This work is really needed so lets move fast on replying to Jamie please.

arodidev: @ebambo  The code itself lives on the form builder, but I do believe now that the core of the logic has been separated, it can be imported from where it is, or stored anywhere where it is accessible into the required use case, maybe some more input from @samuelmale

src/components/interactive-builder/interactive-builder.component.tsx
```diff
@@ -545,6 +562,32 @@ const InteractiveBuilder: React.FC<InteractiveBuilderProps> = ({
                                           questionIndex
                                         )}
                                       </Droppable>
+                                      <br />
+                                        <p
+                                          className={responses.find((item) => item.id === question.id)?.resolution.includes("✅")? styles.approval: styles.validationError}
+                                        >
```
samuelmale, line 568: Do we really need to pitch or flag positive results? Plus in my POV, `resolution.includes("✅")` this isn't the best way of classifying and identifying validation results.

src/form-validator.ts
```diff
@@ -0,0 +1,146 @@
+import { openmrsFetch } from "@openmrs/esm-framework";
+import { async } from "rxjs";
+
+export const handleFormValidation = async (schema) => {
```
samuelmale, line 4: Ideally, this function should just return already processed and classified validation results. As part of the results, I would only expect errors and warnings classified in an explicit way.

src/components/interactive-builder/interactive-builder.component.tsx
```diff
@@ -63,6 +68,14 @@ const InteractiveBuilder: React.FC<InteractiveBuilderProps> = ({
...
   const [showDeleteSectionModal, setShowDeleteSectionModal] = useState(false);
   const [showDeleteQuestionModal, setShowDeleteQuestionModal] = useState(false);
+  const [responses, setResponses] = useState([]);
+
+  const validateForm = () => {
+    handleFormValidation(schema).then(response => {
+      const [questionResolutionsArray, answersResolutionsArray] = response
+      setResponses(questionResolutionsArray)
+    })
+  }
```
samuelmale, line 78: One other way of injecting the validator would be through a hook:

```javascript
// the validator runs if doValidate is set to true
// isValidating shows the validation status
const { errors, warnings, isValidating } = useSchemaValidationResults(doValidate, schema);

```

samuelmale:
@arodidev Did you see the build failure

src/components/form-editor/form-editor.component.tsx
```diff
@@ -163,6 +168,7 @@ const FormEditor: React.FC = () => {
                       schema={schema}
                       onSchemaChange={updateSchema}
                       isLoading={isLoadingFormOrSchema}
+                      validateStateSetter={validationSetter}
```
samuelmale, line 171:
```suggestion
                      validateStateSetter={setIsValidating}
```

src/components/form-editor/form-editor.component.tsx
```diff
@@ -98,6 +99,10 @@ const FormEditor: React.FC = () => {
     localStorage.setItem("formJSON", JSON.stringify(updatedSchema));
   }, []);
 
+  const validationSetter = () => {
+    setIsValidating((prevState) => !prevState);
+  };
+
```
samuelmale, line 105: You don't need this

samuelmale:
@denniskigen can you kindly have a look at this?

src/components/form-editor/form-editor.component.tsx
samuelmale: Actually, I would rename `validateStateSetter` to `setIsValidating`

arodidev: The issues cited have been resolved, kindly review @samuelmale

NethmiRodrigo:
Closing this since its outdated now. Thank you for working on this Jamie! Please feel free to re-open this if need be

PR #271, opened 2024-05-16
(feat) O3-3218: Make the schema validation(block/allow) render a configuration

gitcliff:
## Requirements

## Summary
This PR is addressing the issue enabling on and off schema validation via configuration, also during schema validation the user should be shown the validation errors but still be able to render and saving their changes 
## Screenshots
https://www.loom.com/share/cfc15777ebcf4cc6991110cd06a08638

### DIFF
--- a/src/components/form-editor/form-editor.component.tsx
+++ b/src/components/form-editor/form-editor.component.tsx
@@ -79,6 +79,7 @@ const FormEditor: React.FC = () => {
   const [publishedWithErrors, setPublishedWithErrors] = useState(false);
   const [errors, setErrors] = useState<Array<MarkerProps>>([]);
   const [validationOn, setValidationOn] = useState(false);
+  const { blockRenderingWithErrors } = useConfig();
 
   const isLoadingFormOrSchema = Boolean(formUuid) && (isLoadingClobdata || isLoadingForm);
 
@@ -215,7 +216,6 @@ const FormEditor: React.FC = () => {
 
   const renderSchemaChanges = useCallback(() => {
     resetErrorMessage();
-    setValidationOn(true);
     {
       try {
         const parsedJson: Schema = JSON.parse(stringifiedSchema);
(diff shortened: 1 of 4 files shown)


hadijahkyampeire:
Thanks @gitcliff but I think we should always show the validation errors and highlights, the config should only be attached to the render, whether to allow rendering with errors or not. And the default should be false, then implementations that want to enforce validation will turn it true.

src/config-schema.ts
```diff
@@ -59,4 +59,9 @@ export const configSchema = {
     _default: false,
     _description: 'Whether to enable form validation',
   },
+  enableSchemaValidation: {
```
hadijahkyampeire, line 62: This should be changed to something like `blockRenderingWithErrors` in other words, allowing/not allowing to render if there are errors.

src/components/form-editor/form-editor.component.tsx
```diff
@@ -215,20 +215,17 @@ const FormEditor: React.FC = () => {
...
-    } else {
-      try {
-        const parsedJson: Schema = JSON.parse(stringifiedSchema);
-        updateSchema(parsedJson);
-        setStringifiedSchema(JSON.stringify(parsedJson, null, 2));
-      } catch (e) {
-        if (e instanceof Error) {
-          setInvalidJsonErrorMessage(e.message);
-        }
+
```
hadijahkyampeire, line 218: The config should be used here since it is deciding whether to render or not.

hadijahkyampeire (commented):
Let's tweak the approach, I have left comments to guide in that, let me know if you need more clarification.

src/components/form-editor/form-editor.component.tsx
gitcliff: putting the config here doesn't produced the desired behavior ie `when the configuration is set to true , any change on the schema will fail render the changes ie change of labels and hence can not validate and publish .When set to false it will render the changes but can not validate and publish.`
the rending fails even the config is set to false

gitcliff:
> Thanks @gitcliff but I think we should always show the validation errors and highlights, the config should only be attached to the render, whether to allow rendering with errors or not. And the default should be false, then implementations that want to enforce validation will turn it true.

@hadijahkyampeire the validation errors and highlights are show when the user clicks render their changes . Default is false

src/components/form-editor/form-editor.component.tsx
hadijahkyampeire: okay, that's good but even if it's true, we should still show errors even if it can render. Bottom line, no matter what we need to show schema validation errors when the render button is clicked.

hadijahkyampeire:
So @gitcliff the thing is that the errors showing shouldn't depend on the config, we want to show errors and their highlights even if the config is false, the only thing the config should allow/block is the rendering part.

hadijahkyampeire (approved):
LGTM, thanks @gitcliff , let's merge and see what our users think.

PR #274, opened 2024-05-20
(chore) Bump dependencies

denniskigen:
## Requirements

## Summary

This PR bumps various React and React DOM and their associated typings following [this](https://github.com/openmrs/openmrs-form-engine-lib/pull/278) corresponding version bump in the React Form Engine library. 

Additionally, it bumps various dependencies to their latest compatible minor versions and fixes a type annotation import for our translations tooling.


hadijahkyampeire (dismissed):
LGTM, thanks @denniskigen

denniskigen:
E2E tests failing and I can't test them locally because the dev server keeps reloading because the dev server doesn't ignore changes to the `test-results` directory. Welp, @ibacher.

ibacher (dismissed):

ibacher:
@denniskigen The dev server itself of the playwright tests?

ibacher:
@denniskigen The blocker on the tests seems to be that (with the changes) the tool-tip for "Edit Form" flickers in and out and seems to capture the "click" event, so the tests never get to the page they expect to.

denniskigen:
That's right. So Playwright can't launch a form in Edit mode.

ibacher:
So, [this](https://github.com/openmrs/openmrs-esm-form-builder/pull/274/commits/1245fff85f756957c8af7ab7793025f66b62dfca) fixes things. I don't quite know why... My browser broke while I was trying to figure out what the actual difference was.

Basically, without that change, the tooltip was being rendered in the dead center of the button, so that a cursor placed in the middle of the button would hover over the icon, firing the `onMouseEnter()` event, then the tooltip would render immediately under the cursor, causing the focused element to be the tooltip instead of the button, which in turn, fired the `onMouseExit()` event. Basically, if you left the cursor hovering over the button, it would render and then immediately un-render the tooltip constantly. My guess is that the `cds-popover--top-start` class is either broken or not correctly loaded, but I couldn't confirm that.

ibacher (approved):

PR #317, opened 2024-07-16
(fix) Fix modal header close button styles

denniskigen:
## Requirements

## Summary

This PR implements a temporary fix for the broken styling of the [ComposedModal](https://react.carbondesignsystem.com/?path=/docs/components-composedmodal--overview) component's close button. The issue appears to be caused by a version mismatch in the Carbon Design System.

This patch uses scoped global styles to adjust the close button and its tooltip to approximate the intended design. While not an ideal solution, it ensures the component remains functional and visually consistent for now.

The long-term resolution will involve updating Carbon dependencies across the project to align with the latest version. However, this immediate fix addresses the urgent styling break caused by the recent Carbon version upgrade in this repository.

### Before

https://github.com/user-attachments/assets/8faa28c2-c1aa-46bd-9c4f-a2efbdb63f3a

### After

https://github.com/user-attachments/assets/c00d42df-7935-445d-acc1-72dbcc276cb8

### DIFF
--- a/src/components/action-buttons/action-buttons.scss
+++ b/src/components/action-buttons/action-buttons.scss
@@ -4,7 +4,7 @@
   justify-content: flex-end;
   margin: 1rem 0;
 
-  button {
+  > button {
     margin-left: 1rem
   }
 }
--- a/src/components/action-buttons/unpublish-form.modal.tsx
+++ b/src/components/action-buttons/unpublish-form.modal.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { Button, ModalBody, ModalHeader, ModalFooter } from '@carbon/react';
+import styles from '../modals.scss';
 
 interface UnpublishModalProps {
   closeModal: () => void;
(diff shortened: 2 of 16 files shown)


denniskigen:
@ibacher any tips for better approaching this issue?

ibacher:
My only suggestion here would be that we apply a class to the `ComposedModal` instances so we can scope these style changes to that class, rather than making them global. If that won't work, we should probably move these changes to the styleguide itself so app behaviour remains consistent regardless of whether this app is installed or loaded.

src/components/action-buttons/action-buttons.scss
```diff
@@ -4,7 +4,7 @@
   justify-content: flex-end;
   margin: 1rem 0;
 
-  button {
+  > button {
```
denniskigen, line 7: This ensures that only `button` elements that are direct children of `.actionButton` are affected by these styles.

ibacher (approved):

denniskigen:
Thanks, @ibacher. Let's see how this goes.

PR #336, opened 2024-08-08
(fix) O3-3713: Should not go into infinite loading when conceptId is undefined

NethmiRodrigo:
## Requirements

## Summary

This PR fixes the issue of the backing concept field in the `editQuestionModal` going in to an infinite load if there is no concept or if the concept is set to an empty string.

## Screenshots
Before the fix:

https://github.com/user-attachments/assets/59dac41d-f7fe-4ce7-b0dc-b7ff48df3c37

After the fix:

https://github.com/user-attachments/assets/4c7f1fe3-39be-4b99-b4b9-94afeb66dd13

https://github.com/user-attachments/assets/beeeb34f-6999-4280-8acc-bc5aa5529449

## Related Issue

https://issues.openmrs.org/browse/O3-3713

### DIFF
--- a/src/hooks/useConceptName.ts
+++ b/src/hooks/useConceptName.ts
@@ -1,5 +1,6 @@
 import useSWR from 'swr';
 import { openmrsFetch, restBaseUrl } from '@openmrs/esm-framework';
+import { isEmpty } from '@openmrs/openmrs-form-engine-lib';
 
 export function useConceptName(conceptId: string | undefined) {
   const customRepresentation = 'custom:(name:(display))';
@@ -10,6 +11,6 @@ export function useConceptName(conceptId: string | undefined) {
   return {
     conceptName: data?.data?.name?.display ?? null,
     conceptNameLookupError: error,
-    isLoadingConceptName: (!data && !error) || false,
+    isLoadingConceptName: !conceptId || isEmpty(conceptId) ? false : (!data && !error) || false,
   };
 }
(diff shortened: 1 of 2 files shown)


NethmiRodrigo:
@ibacher @denniskigen @samuelmale I had to update core and framework to get this running but when I do, I get the error about the date picker, so I didn't push yarn.lock
Screenshot 2024-08-08 at 6 45 18 PM

denniskigen:
You need to add this line to the `lib` targets in the tsconfig.json file:

```
"lib": [
	"es2022 "
]
```

src/hooks/useConceptName.ts
```diff
@@ -1,5 +1,6 @@
 import useSWR from 'swr';
 import { openmrsFetch, restBaseUrl } from '@openmrs/esm-framework';
+import { isEmpty } from '@openmrs/openmrs-form-engine-lib';
```
denniskigen, line 3: Did you mean to import from lodash or some other lib? I think a null check would be sufficient here: `!conceptId`

denniskigen (approved):
Thanks, @NethmiRodrigo! LGTM, save for an issue about the `isEmpty` check.

src/hooks/useConceptName.ts
NethmiRodrigo: Done!

denniskigen (approved):
LGTM

PR #343, opened 2024-08-29
(fix): Add min and max values to the JSON schema

NethmiRodrigo:
## Requirements

## Summary

1. When creating a numeric rendering type question using the interactive builder, the `min` and `max` properties were not getting added to the JSON schema. This PR fixes that issue and also shows a warning in the min and max fields if the `min` > `max`.

2. This PR updates the e2e tests to follow standards and adds the missing steps.

## Screenshots

https://github.com/user-attachments/assets/33d6ceb6-2479-41cf-80b4-9cc5537c1c71

## Related Issue

https://issues.openmrs.org/browse/O3-3880


NethmiRodrigo:
One of the E2E tests fail, I don't know why the error `highlightedItem.scrollIntoView is not a function` comes up...

denniskigen:
Thanks for working on this, @NethmiRodrigo! Could we make it so that if I select a concept with min and max values from the concept lookup field, those values automatically get fed into the `min` and `max` input fields if the rendering type is `number` and the field type is `obs`?

NethmiRodrigo:
@denniskigen I did look into it and it might not be straightforward because of the attributes. eg - looking at the concept `Pulse`, either `hi absolute` `hi critical` should come in for the max value, but are these attributes common to all concepts that are of type `Numeric`. Because this might not be the case, we had made this ticket - https://openmrs.atlassian.net/browse/O3-3949
image

src/components/interactive-builder/add-question.modal.tsx
```diff
@@ -484,13 +486,25 @@ const AddQuestionModal: React.FC<AddQuestionModalProps> = ({
                         id="min"
                         labelText="Min"
                         value={min || ''}
+                        invalid={parseInt(min) > parseInt(max)}
```
ibacher, line 489: We don't generally restrict obs to being integers so it may be important not to do so here? (`parseFloat()` vs `parseInt()`.

NethmiRodrigo: Done!

denniskigen:
Disregard my suggestion from earlier, I'd mistakenly assumed that `Numeric` concepts have explicit `min` and `max` values. Thanks for the clarification. LGTM

e2e/specs/interactive-builder.spec.ts
```diff
@@ -12,18 +58,176 @@ test('Create a schema using the interactive builder', async ({ page }) => {
...
+    await formBuilderPage.page.getByRole('option', { name: 'Yes' }).click();
+  });
+
+  await test.step('And then I click on `Save`', async () => {
+    await formBuilderPage.saveQuestionButton().click();
+    await expect(formBuilderPage.page.getByText(/new question created/i)).toBeVisible();
+  });
+
+  await test.step('Then the JSON schema should have the question object', async () => {
+    await context.grantPermissions(['clipboard-read', 'clipboard-write']);
```
denniskigen, line 186: nifty 👩🏼‍🍳

denniskigen (approved):
Excellent work with the tests. Thanks

ibacher:
> I'd mistakenly assumed that Numeric concepts have explicit min and max values.

Where they do (its rare, and only indicated by "Hi Absolute" and "Low Absolute"—the other thresholds are technically clinical guidelines). I actually find it odd that "Pulse" would have absolute values... a pulse cannot be below 0, but there's only a physiologic limit to how high pulse can go. Usually its for things like O2 Saturation, which is a percentage.

PR #347, opened 2024-09-09
(feat) O3-3955: Add question info text field to the interactive builder

andie23:
# Changes
1. Added new type in Question interface
2. Updated Add Question Editor and Edit Question Editor to include question info

### DIFF
--- a/src/const.ts
+++ b/src/const.ts
@@ -1,0 +1,1 @@
+export const QUESTION_PROPS = [{ id: 'questionInfo', text: 'Question Info' }];
--- a/src/types.ts
+++ b/src/types.ts
@@ -117,6 +117,7 @@ export interface Question {
   id: string;
   label: string;
   type: string;
+  questionInfo?: string;
   questionOptions: QuestionOptions;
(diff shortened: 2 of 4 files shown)


src/const.ts
NethmiRodrigo, line 1: I think this is something we can move into the `config-schema.ts`

src/types.ts
```diff
@@ -117,6 +117,7 @@ export interface Question {
   id: string;
   label: string;
   type: string;
+  questionInfo?: string;
```
NethmiRodrigo, line 120: Shall we update the `Schema` interface as well to include `questionInfo`?

NethmiRodrigo (commented):
Great work @andie23!! Just some minor nitpicks. Also, could you please update your PR description to follow the template? It helps a ton when reviewing :)

NethmiRodrigo:
Closing since its inactive, please feel free to re-open this if you begin working on this again. Thanks!

harshthakkr:
@NethmiRodrigo If you’re okay with it, I can create a new PR with the same changes so we can move forward.

PR #351, opened 2024-09-11
 (feat) O3-3712: Add additional concepts as answers to a coded question

Willie-theBeastMutua:
## Requirements

## Summary
Users are now be able to manually add specific concepts as answers to a concept of type Coded after searching of them (similar to how we add a backing concept)
## Screenshots
BeforeClick

AfterClick
select answer 1
select Ans 2
selected ans 3
select ans 4
select ans 5

## Related Issue

https://openmrs.atlassian.net/browse/O3-3712

### DIFF
--- a/src/components/interactive-builder/add-question.modal.tsx
+++ b/src/components/interactive-builder/add-question.modal.tsx
@@ -736,7 +736,7 @@ const AddQuestionModal: React.FC<AddQuestionModalProps> = ({
                       ))}
                     </div>
                   ) : null}
-                  {selectedConcept.datatype?.name === 'Coded' ? (
+                  {selectedConcept && selectedConcept.datatype?.name === 'Coded' ? (
                     <div>
                       <FormLabel className={styles.label}>
                         {t('searchForAnswerConcept', 'Search for a concept to add as an answer')}
--- a/src/components/interactive-builder/add-question.modal.tsx
+++ b/src/components/interactive-builder/add-question.modal.tsx
@@ -107,7 +107,6 @@ const AddQuestionModal: React.FC<AddQuestionModalProps> = ({
   const [isQuestionRequired, setIsQuestionRequired] = useState(false);
   const [max, setMax] = useState('');
   const [min, setMin] = useState('');
-  const [addAnswer, setAnswer] = useState(false);
   const [questionId, setQuestionId] = useState('');
   const [questionLabel, setQuestionLabel] = useState('');
   const [questionType, setQuestionType] = useState<QuestionType | null>(null);
@@ -176,7 +175,6 @@ const AddQuestionModal: React.FC<AddQuestionModalProps> = ({
   const handleConceptSelect = (concept: Concept) => {
     const updatedDatePickerType = getDatePickerType(concept);
     if (updatedDatePickerType) setDatePickerType(updatedDatePickerType);
-    setAnswer(false);
     setConceptToLookup('');
     setSelectedConcept(concept);
     setAnswers(
@@ -330,35 +328,7 @@ const AddQuestionModal: React.FC<AddQuestionModalProps> = ({
         });
       }
     }
-<<<<<<< HEAD
-  }, [
-    onSchemaChange,
-    t,
-    selectedProgram?.uuid,
-    schema,
-    pageIndex,
-    sectionIndex,
-    questionLabel,
-    questionType,
-    isQuestionRequired,
-    questionId,
-    renderingType,
-    datePickerType,
-    selectedConcept,
-    conceptMappings,
-    selectedAnswers,
-    addObsComment,
-    addInlineDate,
-    selectedPersonAttributeType,
-    selectedPatientIdetifierType,
-    selectedProgramState,
-    programWorkflow,
-    toggleLabelTrue,
-    toggleLabelFalse,
-  ]);
-=======
   };
->>>>>>> 1b94b7b (removed callback)
 
   const convertLabelToCamelCase = () => {
     const camelCasedLabel = questionLabel
@@ -368,14 +338,6 @@ const AddQuestionModal: React.FC<AddQuestionModalProps> = ({
       })
       .replace(/\s+/g, '');
     setQuestionId(camelCasedLabel);
-  };
-
-  const showAddQuestion = () => {
-    if (!addAnswer) {
-      setAnswer(true);
-      return;
-    }
-    setAnswer(false);
   };
 
   const handleProgramWorkflowChange = (selectedItem: ProgramWorkflow) => {
@@ -774,17 +736,10 @@ const AddQuestionModal: React.FC<AddQuestionModalProps> = ({
                       ))}
                     </div>
                   ) : null}
-                  {selectedConcept && answers?.length ? (
-                    <div>
-                      <Button kind="tertiary" onClick={showAddQuestion} iconDescription="Add" size="sm">
-                        More Answers
-                      </Button>
-                    </div>
-                  ) : null}
-                  {addAnswer ? (
+                  {selectedConcept.datatype?.name === 'Coded' ? (
                     <div>
                       <FormLabel className={styles.label}>
-                        {t('searchForAnswerConcept', 'Search for an answer Concept to Add')}
+                        {t('searchForAnswerConcept', 'Search for a concept to add as an answer')}
                       </FormLabel>
                       {conceptAnsLookupError ? (
                         <InlineNotification
--- a/src/components/interactive-builder/add-question.modal.tsx
+++ b/src/components/interactive-builder/add-question.modal.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import { useTranslation, type TFunction } from 'react-i18next';
 import flattenDeep from 'lodash-es/flattenDeep';
 import {
@@ -9,7 +9,6 @@ import {
   FormLabel,
   InlineLoading,
   InlineNotification,
-  IconButton,
   Layer,
   ModalBody,
   ModalFooter,
@@ -26,7 +25,7 @@ import {
   TextInput,
   Tile,
 } from '@carbon/react';
-import { ArrowUpRight, Add, TrashCan } from '@carbon/react/icons';
+import { ArrowUpRight } from '@carbon/react/icons';
 import { showSnackbar, useConfig, useDebounce } from '@openmrs/esm-framework';
 import type { ProgramState, RenderType } from '@openmrs/esm-form-engine-lib';
 
@@ -61,10 +60,8 @@ interface AddQuestionModalProps {
 }
 
 interface Item {
-  id: string;
   text: string;
 }
-
 interface ProgramStateData {
   selectedItems: Array<ProgramState>;
 }
@@ -143,6 +140,7 @@ const AddQuestionModal: React.FC<AddQuestionModalProps> = ({
   const [addInlineDate, setAddInlineDate] = useState(false);
   const [selectedProgramState, setSelectedProgramState] = useState<Array<ProgramState>>([]);
   const [selectedProgram, setSelectedProgram] = useState<Program>(null);
+  const [isCreatingQuestion, setIsCreatingQuestion] = useState(false);
   const [programWorkflow, setProgramWorkflow] = useState<ProgramWorkflow>(null);
   const { programs, programsLookupError, isLoadingPrograms } = usePrograms();
   const { programStates, programStatesLookupError, isLoadingProgramStates, mutateProgramStates } = useProgramWorkStates(
@@ -199,28 +197,18 @@ const AddQuestionModal: React.FC<AddQuestionModalProps> = ({
       }),
     );
   };
-  const clearAnsConcept = () => {
-    setaddedAnswers([]);
+  const handleDeleteAnswer = (id) => {
+    setaddedAnswers((prevAnswers) => prevAnswers.filter((answer) => answer.id !== id));
   };
-  const handleSaveClick = () => {
-    setSelectedAnswers((prevAnswers) => {
-      const newAnswers = addedAnswers.filter(
-        (newAnswer) => !prevAnswers.some((prevAnswer) => prevAnswer.id === newAnswer.id),
-      );
-      const updatedAnswers = [...prevAnswers, ...newAnswers];
+  const handleSaveMoreAnswers = () => {
+    const newAnswers = addedAnswers.filter(
+      (newAnswer) => !selectedAnswers.some((prevAnswer) => prevAnswer.id === newAnswer.id),
+    );
 
-      setAnswers((prevAnswers) => [
-        ...prevAnswers,
-        ...newAnswers.map((answer) => ({
-          concept: answer.id,
-          label: answer.text,
-        })),
-      ]);
-
-      return updatedAnswers;
-    });
-
+    const updatedAnswers = [...selectedAnswers, ...newAnswers];
+    setSelectedAnswers(updatedAnswers);
     setaddedAnswers([]);
+    setIsCreatingQuestion(true);
   };
 
   const handleConceptAnsSelect = (concept: Concept) => {
@@ -250,11 +238,11 @@ const AddQuestionModal: React.FC<AddQuestionModalProps> = ({
   };
 
   const handleCreateQuestion = () => {
-    createQuestion();
+    handleSaveMoreAnswers();
     closeModal();
   };
 
-  const createQuestion = () => {
+  const createQuestion = useCallback(() => {
     try {
       const questionIndex = schema.pages[pageIndex]?.sections?.[sectionIndex]?.questions?.length ?? 0;
       const computedQuestionId = `question${questionIndex + 1}Section${sectionIndex + 1}Page-${pageIndex + 1}`;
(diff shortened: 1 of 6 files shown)


NethmiRodrigo:
Can you please add these changes to the `edit-modal` component as well?

Willie-theBeastMutua:
Okay doing that

Willie-theBeastMutua:
Hello @NethmiRodrigo , I added the changes to edit-modal component as requested

package.json
```diff
@@ -48,7 +51,9 @@
   "dependencies": {
     "@carbon/react": "^1.47.0",
     "@openmrs/esm-form-engine-lib": "next",
+    "@openmrs/openmrs-form-engine-lib": "^2.1.0",
     "ajv": "^8.17.1",
+    "cheerio": "^1.0.0",
```
NethmiRodrigo, line 56: I don't see this being used anywhere

src/components/interactive-builder/add-question.modal.tsx
```diff
@@ -19,15 +20,15 @@ import {
...
   Tag,
   TextInput,
   Tile,
+  SelectSkeleton,
 } from '@carbon/react';
-import { ArrowUpRight } from '@carbon/react/icons';
+import { ArrowUpRight, Add, TrashCan } from '@carbon/react/icons';
 import { showSnackbar, useConfig, useDebounce } from '@openmrs/esm-framework';
-import type { ProgramState, RenderType } from '@openmrs/esm-form-engine-lib';
+import type { ProgramState, RenderType } from '@openmrs/openmrs-form-engine-lib';
```
NethmiRodrigo, line 31: Looks like some changes have gotten reverted to a previous state, could you please update your branch and check again?

src/components/interactive-builder/add-question.modal.tsx
```diff
@@ -59,6 +58,14 @@ interface AddQuestionModalProps {
...
 }
 
+const DatePickerType = {
+  both: 'both',
+  calendar: 'calendar',
+  timer: 'timer',
+} as const;
+
+type DatePickerTypeValue = (typeof DatePickerType)[keyof typeof DatePickerType];
+
```
NethmiRodrigo, line 68: These are old too

src/components/interactive-builder/add-question.modal.tsx
```diff
@@ -73,21 +80,6 @@ interface RequiredLabelProps {
...
-    case 'Date':
-      return 'calendar';
-    case 'Time':
-      return 'timer';
-    default:
-      console.warn(`Unsupported datatype for date fields: ${conceptDataType}`);
-      return null;
-  }
-};
-
```
NethmiRodrigo, line 90: Same here

src/components/interactive-builder/add-question.modal.tsx
```diff
@@ -148,18 +158,11 @@ const AddQuestionModal: React.FC<AddQuestionModalProps> = ({
     programState: ['select'],
   };
 
-  // Maps the data type of a concept to a date picker type.
-  const datePickerTypeOptions: Record<string, Array<DatePickerTypeOption>> = {
-    datetime: [{ value: 'both', label: t('calendarAndTimer', 'Calendar and timer'), defaultChecked: true }],
-    date: [{ value: 'calendar', label: t('calendarOnly', 'Calendar only'), defaultChecked: false }],
-    time: [{ value: 'timer', label: t('timerOnly', 'Timer only'), defaultChecked: false }],
-  };
-
```
NethmiRodrigo, line 157: This shouldn't be removed either

src/components/interactive-builder/add-question.modal.tsx
```diff
@@ -148,18 +158,11 @@ const AddQuestionModal: React.FC<AddQuestionModalProps> = ({
...
   const handleConceptChange = (event: React.ChangeEvent<HTMLInputElement>) => setConceptToLookup(event.target.value);
+  const handleAnsConceptChange = (event: React.ChangeEvent<HTMLInputElement>) =>
+    setConceptAnsToLookup(event.target.value);
 
   const handleConceptSelect = (concept: Concept) => {
-    const updatedDatePickerType = getDatePickerType(concept);
-    if (updatedDatePickerType) setDatePickerType(updatedDatePickerType);
```
NethmiRodrigo, line 162: Same here

src/components/interactive-builder/add-question.modal.tsx
```diff
@@ -215,11 +233,10 @@ const AddQuestionModal: React.FC<AddQuestionModalProps> = ({
         type: questionType,
         required: isQuestionRequired,
         id: questionId ?? computedQuestionId,
-        ...((renderingType === 'date' || renderingType === 'datetime') &&
-          datePickerType && { datePickerFormat: datePickerType }),
+        ...(questionType === 'encounterDatetime' && { datePickerFormat: datePickerType }),
         questionOptions: {
           rendering: renderingType,
-          ...(selectedConcept && { concept: selectedConcept?.uuid }),
```
NethmiRodrigo, line 222: And here

src/components/interactive-builder/add-question.modal.tsx
```diff
@@ -311,11 +335,7 @@ const AddQuestionModal: React.FC<AddQuestionModalProps> = ({
 
   return (
     <>
-      <ModalHeader
-        className={styles.modalHeader}
-        title={t('createNewQuestion', 'Create a new question')}
-        closeModal={closeModal}
-      />
+      <ModalHeader title={t('createNewQuestion', 'Create a new question')} closeModal={closeModal} />
```
NethmiRodrigo, line 338: I don't think removing the className is necessary. Maybe a mistake?

src/components/interactive-builder/add-question.modal.tsx
```diff
@@ -546,7 +566,6 @@ const AddQuestionModal: React.FC<AddQuestionModalProps> = ({
                         id="conceptLookup"
                         onClear={() => {
                           setSelectedConcept(null);
-                          setDatePickerType('both');
```
NethmiRodrigo, line 549: Shouldn't be removed

src/components/interactive-builder/add-question.modal.tsx
```diff
@@ -731,35 +870,33 @@ const AddQuestionModal: React.FC<AddQuestionModalProps> = ({
...
-                        .flat()
-                        .map((type) => (
-                          <RadioButton
-                            id={type.value}
-                            checked={datePickerType === type.value}
-                            labelText={type.label}
-                            onClick={() => setDatePickerType(type.value)}
-                            value={type.value}
-                          />
-                        ))}
```
NethmiRodrigo, line 762: Please revert these changes too

src/components/interactive-builder/question-modal.scss
```diff
@@ -1,5 +1,4 @@
 @use '@carbon/colors';
-@use '@carbon/layout';
```
NethmiRodrigo, line 2: This should be reverted

src/components/interactive-builder/question-modal.scss
```diff
@@ -96,7 +95,7 @@
 }
 
 .multiSelect {
-  :global(.cds--list-box__field--wrapper) {
```
NethmiRodrigo, line 99: This is an override for carbon styling so the classname shouldn't be changed

src/components/interactive-builder/question-modal.scss
```diff
@@ -127,31 +126,14 @@
...
-    }
-
-    .cds--popover--left > .cds--popover > .cds--popover-content {
-      transform: translate(-4rem, 0.85rem);
-    }
-
-    .cds--popover--left > .cds--popover > .cds--popover-caret {
-      transform: translate(-3.75rem, 1.25rem);
-    }
-  }
```
NethmiRodrigo, line 156: This should not be removed

NethmiRodrigo:
@Willie-theBeastMutua can you please update your fork and fix the changes, some of the code has gotten reverted, maybe due to a bad rebase

Willie-theBeastMutua:
Hellor @NethmiRodrigo  the code has been updated and fixed kindly check

NethmiRodrigo:
Hey @Willie-theBeastMutua thank you for the progress. One minor detail, I see that we can't remove the answers once added. Can you add support to remove an added answer, and also add a check that you can't add the same concept multiple times as an answer?

Willie-theBeastMutua:
@NethmiRodrigo  support for deletion and checks for duplicate concept answers has been added kindly check

NethmiRodrigo:
@Willie-theBeastMutua could you add like a small `x` icon to the tags so we could remove individual answers, instead of the trash icon removing all answers? We can remove the trash icon. And could you align the add icon button to be aligned with the search?
Screenshot 2024-10-09 at 12 24 24 PM
Also, the additional answers don't get added to the answers key in the form JSON, so when you click save, all the answers that were added gets lost

Willie-theBeastMutua:
@NethmiRodrigo did you have a look

NethmiRodrigo:
Sorry I couldn't re-review it, will do asap. Thanks for pinging me. Meanwhile, can you resolve the conflicts and update your branch?

Willie-theBeastMutua:
@NethmiRodrigo  branch has been updated

Willie-theBeastMutua:
@NethmiRodrigo  can we restart the test as it times out and when i test the same test from my end it should work. Or advice on what is failing....

NethmiRodrigo:
@Willie-theBeastMutua  The e2e test failing is a metadata issue (unrelated to your changes).
One more minor thing, I can add an answer that's already added in the UI. This doesn't get added to the actual object so when I go into edit, it only shows one instance of the answer, but is there something we can do to restrict adding duplicates of the same answer
Screenshot 2024-10-16 at 12 45 50 PM
Screenshot 2024-10-16 at 12 47 28 PM

Willie-theBeastMutua:
@NethmiRodrigo to get you right.. i added the check for duplicates to only save one instance. You  want the same check to be available in the UI?

NethmiRodrigo:
@Willie-theBeastMutua yep, exactly!

Willie-theBeastMutua:
@NethmiRodrigo the checks have been added. Kindly check

NethmiRodrigo:
Could you please update your branch @Willie-theBeastMutua?

NethmiRodrigo (commented):
Thanks @Willie-theBeastMutua, some very minor suggestions

src/components/interactive-builder/add-question.modal.tsx
```diff
@@ -310,6 +386,12 @@ const AddQuestionModal: React.FC<AddQuestionModalProps> = ({
     setSelectedProgram(selectedItem);
     setProgramWorkflows(selectedItem?.allWorkflows);
   };
+  useEffect(() => {
+    if (isCreatingQuestion) {
+      createQuestion();
+      setIsCreatingQuestion(false);
+    }
+  }, [isCreatingQuestion, createQuestion]);
```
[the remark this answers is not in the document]
Willie-theBeastMutua: The use effect is used due to react's state management when combining the added answers and the selected answers. Saving without checking when the state is changed only saves the selected answers and not the update. I tried not using it but react's state management did not allow me. The call back is used to execute also the use effect. They are both part of the same solution

Willie-theBeastMutua:
> Could you please update your branch @Willie-theBeastMutua?

Okay let me do that

src/components/interactive-builder/add-question.modal.tsx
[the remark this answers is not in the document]
NethmiRodrigo: Ok I see. So instead of having a useEffect and useCallback we could change the `handleSaveMoreAnswers` to return the answers list, like so:
```
const handleSaveMoreAnswers = () => {
    const newAnswers = addedAnswers.filter(
      (newAnswer) => !selectedAnswers.some((prevAnswer) => prevAnswer.id === newAnswer.id),
    );

    const updatedAnswers = [...selectedAnswers, ...newAnswers];
    setSelectedAnswers(updatedAnswers);
    setaddedAnswers([]);
    return updatedAnswers;
  };
```
And then in the `handleCreateQuestion`, we pass in the list to the `createQuestion` function, like so:
```
 const handleCreateQuestion = () => {
    const updatedAnswers = handleSaveMoreAnswers();
    createQuestion(updatedAnswers);
    closeModal();
  };
```
And accept change the `createQuestion` function to accept a parameter and then use that param for the object:
```
const createQuestion = (conceptAnswers) => {
    try {
      const questionIndex = schema.pages[pageIndex]?.sections?.[sectionIndex]?.questions?.length ?? 0;
      const computedQuestionId = `question${questionIndex + 1}Section${sectionIndex + 1}Page-${pageIndex + 1}`;

      const newQuestion = {
        label: questionLabel,
        ......,
        ...(conceptAnswers.length && {
            answers: conceptAnswers.map((answer) => ({
              concept: answer.id,
              label: answer.text,
            })),
          }),
      };
    }
};
```
This way we could avoid the useCallback with a massive dependency list

Willie-theBeastMutua:
@NethmiRodrigo  I was able to remove the callback and useEffect according to your suggestion and it worked perfectly. Thank you so much
Kindly check the changes and advice.

src/components/interactive-builder/add-question.modal.tsx
```diff
@@ -199,13 +240,13 @@ const AddQuestionModal: React.FC<AddQuestionModalProps> = ({
...
-
   const handleCreateQuestion = () => {
-    createQuestion();
+    const updatedAnswers = handleSaveMoreAnswers();
+    createQuestion(updatedAnswers);
     closeModal();
   };
 
-  const createQuestion = () => {
+  const createQuestion = (selectedAnswers) => {
```
NethmiRodrigo, line 249: Its better to rename this parameter since we have a state with the same name

NethmiRodrigo: And the update the newQuestion object to use the passed in argument to set the answers

NethmiRodrigo (changes requested):

Willie-theBeastMutua:
@NethmiRodrigo  the change has been made

src/components/interactive-builder/edit-question.modal.tsx
```diff
@@ -809,15 +864,131 @@ const EditQuestionModal: React.FC<EditQuestionModalProps> = ({
...
+                        {selectedAnswers?.map((answer) => (
+                          <Tag className={styles.tag} key={answer?.id} type={'blue'}>
+                            {answer?.text}
                           </Tag>
                         ))}
                       </div>
-                    )}
+                    ) : null}
+
+                    {(selectedConcept || questionToEdit) && questionToEdit?.questionOptions.answers?.length ? (
```
NethmiRodrigo, line 878: Because of the condition `questionToEdit?.questionOptions.answers?.length`, if a user when creating a question doesn't select any answer, they wouldn't be able to add any answers when editing either, and that shouldn't be the case.

Willie-theBeastMutua: @NethmiRodrigo  I found a way to do it kindly check and advise

src/components/interactive-builder/edit-question.modal.tsx
```diff
@@ -875,7 +876,14 @@ const EditQuestionModal: React.FC<EditQuestionModalProps> = ({
                       </div>
                     ) : null}
 
-                    {(selectedConcept || questionToEdit) && questionToEdit?.questionOptions.answers?.length ? (
+                    {!hasConceptChanged && editConceptInfo?.datatype?.uuid == '8d4a48b6-c2cc-11de-8d13-0010c6dffd0f' ? (
```
NethmiRodrigo, line 879: Why do we have a hard-coded concept UUID?

src/components/interactive-builder/edit-question.modal.tsx
NethmiRodrigo: In the add-question-modal, this condition makes sense. In the edit version, if the user has not selected any answer during creation, they should still be able to add answers. What I meant was that we should just remove this condition and show the `Add more answers` if the concept is of the datatype `coded`

src/components/interactive-builder/edit-question.modal.tsx
NethmiRodrigo: @Willie-theBeastMutua please check my comment, just removing the conditional logic is sufficient, and we'll merge this in. Thanks!

src/components/interactive-builder/edit-question.modal.tsx
Willie-theBeastMutua: @NethmiRodrigo how do we check if the datatype is coded? Because even in add question, the check is done to see if the concept has answers there is no check for the datatype

src/components/interactive-builder/edit-question.modal.tsx
Willie-theBeastMutua: > Why do we have a hard-coded UUID?

The UUID id for the coded datatype that is the check

Willie-theBeastMutua: > @Willie-theBeastMutua please check my comment, just removing the conditional logic is sufficient, and we'll merge this in. Thanks!

If I remove the logic, the more answers will display even when the concept is not of a coded datatype so I added a hook that collects data on the concept and I check for the UUID for the coded datatype. I can use the word "Coded" to check but my fear was what would happen if the language changed

src/components/interactive-builder/edit-question.modal.tsx
NethmiRodrigo: So in the `add-question-modal`, the reason that there is no check for the datatype is because as soon as you search for a concept from the API and select a concept, if the concept has answers, it got added to a state, so the check for the datatype `coded` wasn't necessary (hacky way to do it) - https://github.com/openmrs/openmrs-esm-form-builder/blob/5154c460ff533c4bec4180486e92cc5bda168b05/src/components/interactive-builder/add-question.modal.tsx#L160-L170

src/components/interactive-builder/edit-question.modal.tsx
NethmiRodrigo: Ignore my suggestion about simply removing the condition, I see why the hook is needed. 
@ibacher if we do a datatype.display === "Coded", would that be language sensitive?

NethmiRodrigo: @Willie-theBeastMutua for now lets assume it wont change with the language

ibacher: @NethmiRodrigo Yes, but if you used `datatype.name` is shouldn't be, except in some edge-cases.

ibacher: In general `display` is the (possibly) localized metadata, but `name` is the "raw" value.

NethmiRodrigo (approved):
Thanks @Willie-theBeastMutua, great work! Sorry about the delay and the work that this required! I added some minor nitpicks - opted to remove the button that says `Add answers` and directly showed the search bar to search for more answers and updated the branch

Willie-theBeastMutua:
@NethmiRodrigo  thank you so much. It was a great experience and I hope we can build more in the future

PR #354, opened 2024-10-07
(feat) O3-3947:  Add support for different question types

Kinotijoan:
## Requirements

## Summary

I have added support for some question types that were in the json-schema but not in the interactive builder  hence I have finalised on this ticket

## Screenshots
image

image

image

image

image

image

image

## Related Issue
https://openmrs.atlassian.net/browse/O3-3947


NethmiRodrigo:
@Kinotijoan can you resolve the merge conflicts and update your fork?

NethmiRodrigo:
For questions of type `markdown` we don't need a label, if we could conditionally not render it that would be better because otherwise it can be misleading

Kinotijoan:
> For questions of type `markdown` we don't need a label, if we could conditionally not render it that would be better because otherwise it can be misleading

sure

src/components/interactive-builder/edit-question.modal.tsx
```diff
@@ -149,8 +149,11 @@ const EditQuestionModal: React.FC<EditQuestionModalProps> = ({
     questionToEdit?.questionOptions.concept &&
     questionToEdit?.questionOptions?.concept !== selectedConcept?.uuid;
   const [addInlineDate, setAddInlineDate] = useState(false);
-  const [toggleLabelTrue, setToggleLabelTrue] = useState(questionToEdit?.questionOptions?.toggleOptions?.labelTrue);
-  const [toggleLabelFalse, setToggleLabelFalse] = useState(questionToEdit?.questionOptions?.toggleOptions?.labelFalse);
+  const [toggleLabelTrue, setToggleLabelTrue] = useState('');
+  const [toggleLabelFalse, setToggleLabelFalse] = useState('');
```
NethmiRodrigo, line 153: This should be reverted

src/components/interactive-builder/edit-question.modal.tsx
```diff
@@ -149,8 +149,11 @@ const EditQuestionModal: React.FC<EditQuestionModalProps> = ({
...
+  const [formMarkdown, setFormMarkdown] = useState([]);
+  const [buttonLabel, setButtonLabel] = useState('');
+  const [workspaceName, setWorkspaceName] = useState('');
```
NethmiRodrigo, line 156: These variables should be initiated not with an empty string but with the relevant values from the `questionToEdit` object

PR #355, opened 2024-10-14
(feat) O3-3946: Add support for markdown questions

Twiineenock:
## Requirements

### Description

This pull request introduces support for creating markdown-based questions within the interactive form builder. Markdown questions are designed for scenarios where additional explanatory content is needed within the form, apart from standard question labels. This content can help users understand how to fill out fields, explain the meaning of specific fields, or add other context. Since these questions are meant for **accessibility** and **enhanced guidance**, they do not contain fields to fill out.

> **Note**: This PR complements the work in the `form-engine` repository. See [feat: Tweak the markdown support #411](https://github.com/openmrs/openmrs-esm-form-engine-lib/pull/411) for additional details on markdown support improvements related to this update.

### Implementation Details

Currently, there is markdown rendering support in the `openmrs-esm-form-engine-lib` using [react-markdown](https://www.npmjs.com/package/react-markdown), which renders markdown strings as HTML elements within the UI. 

To provide full markdown support with editing capabilities, this PR includes the following key components:

- **Markdown Editor UI**: Integrated using [react-mde](https://www.npmjs.com/package/react-mde), allowing users to easily create and edit markdown content.
- **Markdown Renderer**: Uses [react-markdown](https://www.npmjs.com/package/react-markdown) to safely render markdown content to React elements, supporting common markdown features such as:
  - **Headings**
  - **Bold and Italics**
  - **Lists** (ordered, unordered, and task lists)
  - **Strikethrough**
  - **Superscript and Subscript**

### Key Changes

1. **UI Enhancements**:
   - Added a markdown editor UI in the form builder, allowing users to create and edit markdown-based questions directly.
   - Ensured a rich text editing experience with full markdown support for common styling and structuring elements.

2. **Integration with `form-engine`**:
   - Minor adjustments within the `openmrs-esm-form-engine-lib` to seamlessly integrate the markdown editor with existing question types.
   - Leveraged `react-markdown` within the form engine to convert markdown text into React-rendered HTML elements.

### Features Added in this PR

- **Creation of New Markdown Questions**: Users can now create new markdown-based questions, formatted using markdown syntax.
- **Editing Existing Markdown Questions**: Markdown questions can be easily edited, providing flexibility in adding or updating content.

By implementing this feature, the form builder now offers greater flexibility and accessibility, allowing clients to incorporate rich text explanations and formatting within their forms.

**Technical Note**: The core of this markdown support relies on `react-markdown` for rendering and `react-mde` for user editing. These libraries handle markdown safely and ensure React compatibility.

### Links
- [react-markdown Documentation](https://www.npmjs.com/package/react-markdown)
- [react-mde Documentation](https://www.npmjs.com/package/react-mde)
- [Form Engine Repository](https://github.com/openmrs/openmrs-esm-form-engine-lib)

## Screenshots

form-preview

interactive-builder

preview

## Screancasts
Creating a new question:

[creat.webm](https://github.com/user-attachments/assets/1b5e4b76-96a5-4d37-9935-7a41e0e33532)

Edit an existing Question

[edit-qn.webm](https://github.com/user-attachments/assets/fbc9491b-a625-43e1-9183-f23cc5e1caaa)

## Related Issue

[https://openmrs.atlassian.net/browse/O3-3946](https://openmrs.atlassian.net/browse/O3-3946)


NethmiRodrigo:
@Twiineenock could you please update your branch?

src/components/interactive-builder/draggable-question.component.tsx
```diff
@@ -89,7 +104,9 @@ const DraggableQuestion: React.FC<DraggableQuestionProps> = ({
             <Draggable />
           </IconButton>
         </div>
-        <p className={styles.questionLabel}>{question.label}</p>
+        <p className={styles.questionLabel}>
+          <MarkdownWrapper markdown={question.label || question.value} />
+        </p>
```
NethmiRodrigo, line 109: Since the text shown in markdown can be quite long, unlike a question label, I don't think we should show the entirety of it here. Maybe we could just say `Markdown content`, or something along those lines?
@denniskigen thoughts?

src/components/interactive-builder/markdown-question.component.tsx
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
```
NethmiRodrigo, line 1: Love that you made this into a component of its own! Thank you!

src/components/interactive-builder/edit-question.modal.tsx
```diff
@@ -372,13 +375,28 @@ const EditQuestionModal: React.FC<EditQuestionModalProps> = ({
...
-              required
-            />
+            <Select
+                defaultValue={questionToEdit.questionOptions.rendering}
+                onChange={(event: React.ChangeEvent<HTMLSelectElement>) => setFieldType(event.target.value as RenderType)}
+                id="renderingType"
+                invalidText={t('validFieldTypeRequired', 'A valid field type value is required')}
+                labelText={t('renderingType', 'Rendering type')}
+                required
+              >
```
NethmiRodrigo, line 385: I'd prefer to have the question label be the first input in the modal, it feels more intuitive.

src/components/interactive-builder/markdown-question.component.tsx
```diff
@@ -0,0 +1,54 @@
...
+import ReactMde, { getDefaultToolbarCommands } from 'react-mde';
+import ReactMarkdown from 'react-markdown';
+import remarkGfm from 'remark-gfm';
+import rehypeRaw from 'rehype-raw';
+import styles from './markdown-question.scss';
+
+interface MarkdownQuestionProps {
+  palceHolder?: string;
```
NethmiRodrigo, line 9:
```suggestion
  placeholder?: string;
```

package.json
```diff
@@ -54,6 +54,10 @@
     "fuzzy": "^0.1.3",
     "lodash-es": "^4.17.21",
     "react-ace": "^11.0.1",
+    "react-markdown": "^9.0.1",
+    "react-mde": "^11.5.0",
+    "rehype-raw": "^7.0.0",
+    "remark-gfm": "^4.0.0",
```
NethmiRodrigo, line 60: Are these libraries to support for html tags? If so, I'd rather we leave them out and now have support for HTML tags

NethmiRodrigo: Because I'm concerned about XSS - https://www.npmjs.com/package/react-mde#xss-concerns

NethmiRodrigo: My bad, I see we already support some tags in the form engine - https://github.com/openmrs/openmrs-esm-form-engine-lib/blob/810fd93961be3a38c3e3afe351996d676de65eec/src/components/inputs/markdown/markdown-wrapper.component.tsx#L9

Twiineenock: > Are these libraries to support for html tags? If so, I'd rather we leave them out and not have support for HTML tags

@NethmiRodrigo , we do need `remark-gfm` if we are to have **GitHub Flavored Markdown (GFM)** e.g tables, task lists, strikethroughs, Autolinks e.t.c

And we need The `rehypeRaw` plugin processes raw HTML embedded within Markdown content, allowing it to render as HTML elements in the final output.

Are there some of these formatting options we don't need so we can keep it as simple as possible?

NethmiRodrigo: Lets limit the formatting options to only the ones that are *currently* supported by the react form engine. We don't need these libraries then right?

Twiineenock: > Lets limit the formatting options to only the ones that are _currently_ supported by the react form engine. We don't need these libraries then right?

We now have this implemented

src/components/interactive-builder/draggable-question.component.tsx
Twiineenock: @NethmiRodrigo , we have come up with this:

short-mark-down

If the markdown is very long, we show a small part of the markdown plus ellipsis to show that there is more content.

NethmiRodrigo: Lovely! Thanks @Twiineenock!

Twiineenock:
Hi @NethmiRodrigo, @denniskigen,

I'm currently trying to import the [react-mde](https://www.npmjs.com/package/react-mde#styling) styles directly from the library instead of rewriting them. However, I noticed that our Webpack configuration is enabling CSS Modules globally, which is likely scoping the CSS classes to components, making them unavailable for global styles like those from `react-mde`.

Here’s the relevant part of the Webpack configuration:

```js
const cssLoader = {
    loader: require.resolve('css-loader'),
    options: {
      modules: {
        localIdentName: `${ident}__[name]__[local]___[hash:base64:5]`,
      },
    },
  };
```

I suspect that this configuration is preventing the` react-mde` styles from being applied globally. Could this be the reason why the styles from `react-mde` aren't working as expected?

How can we tweak the Webpack config to ensure that CSS Modules are applied to our local styles, but exclude the styles from `react-mde` (or any other library in `node_modules`) so that they are applied globally without scoping?

Thanks!

NethmiRodrigo:
I can't seem to figure out why the e2e tests are failing, they pass perfectly fine locally. When creating a question, the save button in the modal stays disabled for some reason. Pinging @denniskigen for help

NethmiRodrigo:
> Hi @NethmiRodrigo, @denniskigen,
> 
> I'm currently trying to import the [react-mde](https://www.npmjs.com/package/react-mde#styling) styles directly from the library instead of rewriting them. However, I noticed that our Webpack configuration is enabling CSS Modules globally, which is likely scoping the CSS classes to components, making them unavailable for global styles like those from `react-mde`.
> 
> Here’s the relevant part of the Webpack configuration:
> 
> ```js
> const cssLoader = {
>     loader: require.resolve('css-loader'),
>     options: {
>       modules: {
>         localIdentName: `${ident}__[name]__[local]___[hash:base64:5]`,
>       },
>     },
>   };
> ```
> 
> I suspect that this configuration is preventing the` react-mde` styles from being applied globally. Could this be the reason why the styles from `react-mde` aren't working as expected?
> 
> How can we tweak the Webpack config to ensure that CSS Modules are applied to our local styles, but exclude the styles from `react-mde` (or any other library in `node_modules`) so that they are applied globally without scoping?
> 
> Thanks!

@Twiineenock about this, with help from @ibacher, I changed the webpack config to overwrite the css rule, which is

NethmiRodrigo:
@Twiineenock lets opt to leave the styling as you've done. Although for future work, we should make the markdown editor match the carbon design system - https://carbondesignsystem.com/patterns/text-toolbar-pattern/. All that's left is figuring out why the e2e test isn't working. I doubt it has anything to do with your changes because they pass locally

Twiineenock:
> @Twiineenock lets opt to leave the styling as you've done. Although for future work, we should make the markdown editor match the carbon design system - https://carbondesignsystem.com/patterns/text-toolbar-pattern/. All that's left is figuring out why the e2e test isn't working. I doubt it has anything to do with your changes because they pass locally

Woow, great Idea @NethmiRodrigo ,
Could I apply [carbon design system](https://carbondesignsystem.com/patterns/text-toolbar-pattern/) to the markdown editor before we could let this in?

NethmiRodrigo:
@Twiineenock if you would like to do it in this PR, we can go with that approach. I would prefer we get this in and do that work incrementally, since this feature works perfectly fine as is. About the e2e test, could you please update the `save` button in the add question and edit question modal to remove the check for the label to be enabled, since we made it optional? I believe that is what could be why the e2e test fails

Twiineenock:
> @Twiineenock if you would like to do it in this PR, we can go with that approach. I would prefer we get this in and do that work incrementally, since this feature works perfectly fine as is. About the e2e test, could you please update the `save` button in the add question and edit question modal to remove the check for the label to be not null for the button to be enabled, since we made it optional? I believe that is what could be why the e2e test fails

Alright we shall do that incrementally

NethmiRodrigo:
Just a few final changes and we can merge this in @Twiineenock, could you also please update your branch?

src/types.ts
```diff
@@ -117,7 +117,7 @@ export interface Section {
 export interface Question {
   id: string;
   label?: string;
-  value?:string;
+  value?: string;
   type: string;
```
NethmiRodrigo, line 121: We should make this optional since markdown content doesn't have a type

NethmiRodrigo: And we should update this in form engine too

Twiineenock: @NethmiRodrigo , we have a PR here [feat: FormField Type should be Optioanl #419](https://github.com/openmrs/openmrs-esm-form-engine-lib/pull/419) for this.

src/types.ts
```diff
@@ -117,8 +117,8 @@ export interface Section {
 export interface Question {
   id: string;
   label?: string;
-  value?:string;
-  type: string;
+  value?: string;
```
NethmiRodrigo, line 120: @Twiineenock I see `value` is of type `any` in form-engine and not `string`

Twiineenock: @NethmiRodrigo I had set the type to string because of this `eslint` error `Unexpected any. Specify a different type.eslint[@typescript-eslint/no-explicit-any](https://typescript-eslint.io/rules/no-explicit-any)`

Twiineenock: Should we go for `any` or `string`?

Twiineenock: No worries! Added `/ eslint-disable-next-line @typescript-eslint/no-explicit-any` to escape `eslint` check for `any` when validating question `value` attribute.

src/types.ts
```diff
@@ -117,8 +118,9 @@ export interface Section {
 export interface Question {
   id: string;
   label?: string;
-  value?:string;
-  type: string;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  value?: any;
+  type?: string;
```
NethmiRodrigo, line 123: We should revert making type to optional. That's probably why the build fails

src/types.ts
```diff
@@ -63,8 +63,9 @@ export interface Schema {
       questions: Array<{
         id: string;
         label?: string;
-        value?:string;
-        type: string;
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
NethmiRodrigo, line 66: Rather than adding this as a comment, lets follow what has been done in the form engine and add this as a rule to the eslintrc file. See https://github.com/openmrs/openmrs-esm-form-engine-lib/blob/7f9a0813b1c69450e2cc0fcfa5222cd880b6033d/.eslintrc#L16

NethmiRodrigo (changes requested):

src/types.ts
Twiineenock: Thanks, @NethmiRodrigo! Adding it directly to the `.eslintrc` file, like in the form engine, was a great tip—appreciate the help!

NethmiRodrigo (approved):

NethmiRodrigo:
Alrighty! Great work @Twiineenock! Could you update your branch and we'll merge this in!

src/types.ts
ibacher: Actually, we would prefer that the no-explicit-any rule is _enabled_ and disclaimed like this where it's unavoidable (although, for preference, use `unknown` rather than `any`).

ibacher: FWIW, the only reason its currently disabled in other repos is because historically we didn't enforce it.

src/components/interactive-builder/draggable-question.component.tsx
```diff
@@ -89,7 +129,9 @@ const DraggableQuestion: React.FC<DraggableQuestionProps> = ({
             <Draggable />
           </IconButton>
         </div>
-        <p className={styles.questionLabel}>{question.label}</p>
+        <p className={styles.questionLabel}>
+          {question.label ? question.label : <MarkdownWrapper markdown={question.value} />}
+        </p>
```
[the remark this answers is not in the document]
Twiineenock: @ibacher , does this imply that question type `value` can have other `renderingType`s apart from markdown?

jest.config.js
```diff
@@ -14,6 +14,7 @@ module.exports = {
     'lodash-es': 'lodash',
     '^dexie$': '<rootDir>/node_modules/dexie',
     '^react-i18next$': '<rootDir>/__mocks__/react-i18next.js',
+    'react-markdown': '<rootDir>/__mocks__/react-markdown.tsx',
```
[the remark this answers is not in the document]
NethmiRodrigo: This is a very bad response from me to say this as the reason, but I did this following what we had done in form-engine, because jest threw an error.

src/components/interactive-builder/draggable-question.component.tsx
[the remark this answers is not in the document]
NethmiRodrigo: Maybe in the future, probably. It would be better to check the rendering type === markdown as a condition.

src/components/interactive-builder/edit-question.modal.tsx
```diff
@@ -397,56 +401,60 @@ const EditQuestionModal: React.FC<EditQuestionModalProps> = ({
...
-              invalidText={t('typeRequired', 'Type is required')}
-              labelText={t('questionType', 'Question type')}
-              required
-            >
-              {!questionType && <SelectItem text={t('chooseQuestionType', 'Choose a question type')} value="" />}
-              {questionTypes.map((questionType, key) => (
-                <SelectItem text={questionType} value={questionType} key={key} />
-              ))}
-            </Select>
+            {questionToEdit.questionOptions.rendering !== 'markdown' && (
```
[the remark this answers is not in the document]
NethmiRodrigo: @ibacher I think it might be easier to add this into the refactor

src/components/interactive-builder/draggable-question.component.tsx
```diff
@@ -69,6 +70,45 @@ const DraggableQuestion: React.FC<DraggableQuestionProps> = ({
     }
   }, [handleDuplicateQuestion, isDragging, question, pageIndex, sectionIndex]);
 
+  const MarkdownWrapper: React.FC<{ markdown: string | Array<string> }> = ({ markdown }) => {
```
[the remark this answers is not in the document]
Twiineenock: Updated this!

src/components/interactive-builder/edit-question.modal.tsx
```diff
@@ -397,57 +401,63 @@ const EditQuestionModal: React.FC<EditQuestionModalProps> = ({
...
-              invalidText={t('validFieldTypeRequired', 'A valid field type value is required')}
-              labelText={t('renderingType', 'Rendering type')}
-              required
-            >
-              {!fieldType && <SelectItem text={t('chooseRenderingType', 'Choose a rendering type')} value="" />}
-              {fieldTypes.map((fieldType, key) => (
-                <SelectItem text={fieldType} value={fieldType} key={key} />
-              ))}
-            </Select>
+            {'label' in questionToEdit && (
```
NethmiRodrigo, line 404: Since every question has a label, although these fields are not required for markdown questions, they would get rendered, which doesn't happen in add-question-modal. We should change this condition to match the one we used in the add-question-modal
Screenshot 2024-11-18 at 11 39 14 AM
Screenshot 2024-11-18 at 11 41 04 AM

src/components/interactive-builder/edit-question.modal.tsx
```diff
@@ -397,57 +401,63 @@ const EditQuestionModal: React.FC<EditQuestionModalProps> = ({
...
+                  required
+                >
+                  {!questionType && <SelectItem text={t('chooseQuestionType', 'Choose a question type')} value="" />}
+                  {questionTypes.map((questionType, key) => (
+                    <SelectItem text={questionType} value={questionType} key={key} />
+                  ))}
+                </Select>
+              </>
+            )}
+            {!questionToEdit.value &&
```
NethmiRodrigo, line 446: Why are we conditionally displaying the 'renderingType' field only when it has no value? If the intent is to prevent users from modifying the rendering type, we don't need to do that. Let's always display this field regardless.

NethmiRodrigo:
Thanks for the great work @Twiineenock! Sorry that it took so long to get this in!

Twiineenock:
Thanks for the support @NethmiRodrigo , @ibacher

PR #359, opened 2024-11-25
(refactor) O3-4145: Break question modal into individual components

NethmiRodrigo:
## Requirements

## Summary

This PR aims to refactor the question modal and organize the code structure.

### 1. Refactor to the question modal

**Why we need this?**
The existing implementation had a major problem - there were two separate components (AddQuestionModal and EditQuestionModal) to support creating and editing a question. This meant that 
- a large amount of functionality was duplicated in the two components
- the two components weren't similar in functionality, some things that you could do when creating a question couldn't be done when editing a question (EditQuestionModal was missing functionality that the AddQuestionModal had)
- if we wanted to add more functionality, we'd have to keep on duplicating it
- the way the component was structured made it difficult to add support for ObsGroup questions because in ObsGroup you'd have a question within a question and since we had everything bundled together in either AddQuestionModal or EditQuestionModal, this would require an unnecessarily complicated (and very inefficient) solution
- we didn't have unit tests (because of the lack of units, lol)

**Solution implemented**
Question modal refactor
For both creating and editing a question, the same question modal will be used. All components introduced have at least basic unit tests introduced.
1. Question Modal - contains all the modal content, along with being the parent component for the `Question` component. If the main question is an `ObsGroup`, it will render an option to create a grouped question that will render the `Question` component again.
_Adding support for ObsGroup wasn't meant to be a part of the refactor, but I included it to make sure that the refactor was done in a way that would support this_
Screenshot 2025-01-08 at 5 58 45 PM
Screenshot 2025-01-08 at 5 59 11 PM

2. Question - houses all the input fields that are common to all questions (e.g. - question id) in addition to `RenderingType` component and `QuestionType` component.
4. `QuestionType` - each question type has its own component (e.g - obs) . This makes it easier to add support to new question types without making a singular component too big or complicated. There is a parent `QuestionType` component that has a component map which renders the respective component based on the question type set in the question object.
6. `RenderingType` - similar to `QuestionType` except for rendering types (e.g. -date, number, text).
8. FormFieldContext - Since all the components above need to make changes to the same `question` (or `formfield`) object, a context is used that is provided in the `QuestionModal` and used in all the children components. A context was used because otherwise I would have prop drilled down a state.

* **_Separate from the component refactor, the types were updated to use the ones imported from the `form-engine-lib` and not the ones typed within the form builder itself._** 

### 2. Adjust the code structure
1. All components relevant to the interactive-builder were moved into the `interactive-builder` folder.
2. All modals used were moved into `interactive-builder/modal`
3. All tools were moved into `/tools`

### 3. Additional changes
Added aliases to the folders hooks, resources, and tools and to the types and constants file to avoid long relative paths -  these were added to the webpack config and jest config.

## Related Issue

https://issues.openmrs.org/browse/O3-4145
https://issues.openmrs.org/browse/O3-3950

## Other

Work left to do:
1. Remove usage of explicit types that are defined in `form-engine-lib`. Most types were re-defined within form builder, and are heavily outdated. We should change it to directly import the types from there instead.
2. The unit tests could probably be better written according to standards, and could cover more scenarios.
3. Add support for deleting a sub question of an ObsGroup type question.


src/constants.ts
```diff
@@ -0,0 +1,54 @@
...
+  'encounterDatetime',
+  'encounterLocation',
+  'encounterProvider',
+  'encounterRole',
+  'obs',
+  'obsGroup',
+  'patientIdentifier',
+  'testOrder',
+  'programState',
+];
```
ibacher, line 14: You can get better types this way, I think:

```ts
export const questionTypes = [
  'control',
  'encounterDatetime',
  'encounterLocation',
  'encounterProvider',
  'encounterRole',
  'obs',
  'obsGroup',
  'patientIdentifier',
  'testOrder',
  'programState',
] as const;

export type QuestionType = typeof questionTypes[number];

export const renderTypeOptions: Record<Exclude<QuestionType, 'obs'>, Array<RenderType>> = {
  control: ['text', 'markdown'],
  encounterDatetime: ['date', 'datetime'],
  encounterLocation: ['ui-select-extended'],
  encounterProvider: ['ui-select-extended'],
  encounterRole: ['ui-select-extended'],
  obsGroup: ['group', 'repeating'],
  testOrder: ['group', 'repeating'],
  patientIdentifier: ['text'],
  programState: ['select'],
};
```

Now `QuestionType` can be used where you need to reference a member of `questionTypes` and will be kept in sync with that list.

ibacher (approved):

PR #368, opened 2024-12-06
(chore) Update tooling

NethmiRodrigo:
## Requirements

## Summary

This PR adds the property `skipDefaultValue` to the i18next-parser config because it wasn't picking up changes to the default string without it. It also removes the OHRI script rule config from the webpack config.


NethmiRodrigo:
Unrelated to this specific PR but I don't know why the deployment workflow on main keeps failing saying maximum number of jobs reached every time I re-run it @ibacher

ibacher:
> keeps failing saying maximum number of jobs reached

So on Bamboo, we only allow each job to have a single concurrent run, because of resource limits. What happens is that each merged PR, after it builds, triggers the Bamboo build. If Bamboo is still building the frontend (which takes ~5 minutes), that attempt to trigger the Bamboo job fails. If you're merging a bunch of PRs in one go, as long as the deploy works on the _last_ PR you should be fine (the Bamboo build pulls down whatever is tagged as `latest`, so it will pick up any changes since the last build).

ibacher (approved):

PR #387, opened 2025-02-03
(fix) O3-4424: Fix Context Provider State Merge for Obs Group Questions

Bharath-K-Shetty:
## Requirements

## Summary

This PR addresses an issue with editing obs group questions where changes to the Unique ID and Question Type fields were not persisting. The problem was traced to the context provider, which did not merge updates properly into both the top-level form field state and its nested questions array.

In this update, the FormFieldContext provider is modified so that when a subquestion is updated, its changes are merged into the main form field as well as into the corresponding entry in the questions array. This ensures that users can successfully edit the Unique ID and Question Type fields without the input resetting or only accepting a single character.

## Screenshots

Before

https://github.com/user-attachments/assets/37b6792c-e800-4c33-9ba5-fe7a6ebb06ee

After

https://github.com/user-attachments/assets/01764392-16ef-4ad9-b28f-42b30eef6748

## Related Issue
https://openmrs.atlassian.net/browse/O3-4424

### DIFF
--- a/src/components/interactive-builder/modals/question/form-field-context.tsx
+++ b/src/components/interactive-builder/modals/question/form-field-context.tsx
@@ -23,16 +23,42 @@ export const FormFieldProvider: React.FC<{
   const updateObsGroupedQuestion = useCallback(
     (updatedObsGroupFormField: FormField) => {
       setFormField((prevFormField) => {
-        const formFieldCopy = { ...prevFormField };
-        if (formFieldCopy.questions) {
-          if (formFieldCopy.questions?.length === 1 && formFieldCopy.questions[0].id === '') {
-            formFieldCopy.questions[0] = updatedObsGroupFormField;
-          } else {
-            formFieldCopy.questions.pop();
-            formFieldCopy.questions.push(updatedObsGroupFormField);
-          }
+        let newFormField = { ...prevFormField, ...updatedObsGroupFormField };
+
+        if (!prevFormField.questions) {
+          newFormField.questions = [updatedObsGroupFormField];
+          return newFormField;
         }
-        return formFieldCopy;
+
+        const placeholderIndex = prevFormField.questions.findIndex((q) => q.id === '');
+        if (placeholderIndex !== -1) {
+          const updatedQuestions = [...prevFormField.questions];
+          updatedQuestions[placeholderIndex] = {
+            ...updatedQuestions[placeholderIndex],
+            ...updatedObsGroupFormField,
+          };
+          newFormField.questions = updatedQuestions;
+          return newFormField;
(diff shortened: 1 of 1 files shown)


Bharath-K-Shetty:
@NethmiRodrigo review needed..!

NethmiRodrigo (commented):
@Bharath-K-Shetty Thanks for the attempt! Unfortunately, this doesn't resolve the core issue.

The main problem is that the context providers are nested inside the QuestionModal. My initial mistake was assuming that the nested provider would have access to the entire formField object, but in reality, it only receives the specific child question it is relevant to.

For example, given this formField structure:

```
const formField = {
  id: 'id',
  questions: [{ id: 'child question' }]
};
```
Currently, the child question's context only receives formField.questions[0], meaning it lacks access to the full formField object. However, when the child question is updated, we need to ensure that the parent context’s state is also updated accordingly with the child being updated in the parent state's correct element in the questions array. Basically, we need to push the updates to the parent context. If you test out your code and check the JSON schema, you'll see that the questions array will always be empty values :
`{
              "type": "obsGroup",
              "questionOptions": {
                "rendering": "group"
              },
              "id": "contactedSuccessfully",
              "label": "Contacted Successfully",
              "questions": [
                {
                  "type": "",
                  "id": ""
                }
              ],
              "required": true
            }`

UNCANNY69:
> @Bharath-K-Shetty Thanks for the attempt! Unfortunately, this doesn't resolve the core issue.
> 
> The main problem is that the context providers are nested inside the QuestionModal. My initial mistake was assuming that the nested provider would have access to the entire formField object, but in reality, it only receives the specific child question it is relevant to.
> 
> For example, given this formField structure:
> 
> ```
> const formField = {
>   id: 'id',
>   questions: [{ id: 'child question' }]
> };
> ```
> 
> Currently, the child question's context only receives formField.questions[0], meaning it lacks access to the full formField object. However, when the child question is updated, we need to ensure that the parent context’s state is also updated accordingly with the child being updated in the parent state's correct element in the questions array. Basically, we need to push the updates to the parent context. If you test out your code and check the JSON schema, you'll see that the questions array will always be empty values : `{ "type": "obsGroup", "questionOptions": { "rendering": "group" }, "id": "contactedSuccessfully", "label": "Contacted Successfully", "questions": [ { "type": "", "id": "" } ], "required": true }`

Fixed the addressed issue here

[O3-4424: Fixing the missing context questions in the obsGrouped type by UNCANNY69 · Pull Request #388 · openmrs/openmrs-esm-form-builder](https://github.com/openmrs/openmrs-esm-form-builder/pull/388)

PR #391, opened 2025-02-10
(fix)O3-4320: Remove Irrelevant Properties When Changing Question Type

Bharath-K-Shetty:
## Requirements

## Summary

This PR implements functionality to clean up form field data when a user changes a question's type in the form builder. A mapping of allowed properties for each question type is defined, and a helper function (cleanFormFieldForType) ensures that only the properties relevant to the newly selected question type are retained. The solution also cleans the nested questionOptions object based on a separate mapping, ensuring required keys like "rendering" remain intact. 

## Screenshots

https://github.com/user-attachments/assets/3019d508-62e7-4bf9-90b3-1169008973e6

## Related Issue

https://openmrs.atlassian.net/browse/O3-4320

### DIFF
--- a/src/components/interactive-builder/modals/question/question.modal.tsx
+++ b/src/components/interactive-builder/modals/question/question.modal.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback } from 'react';
+import React, { useCallback, useEffect } from 'react';
 import { useTranslation } from 'react-i18next';
 import flattenDeep from 'lodash-es/flattenDeep';
 import {
@@ -29,6 +29,75 @@ interface QuestionModalProps {
   resetIndices: () => void;
 }
 
+/**
+ * Mapping of allowed top‑level property keys for each question type.
+ * Adjust these keys as needed for your implementation.
+ */
+const allowedPropertiesMapping: Record<string, string[]> = {
+  control: ['id', 'label', 'type', 'questionOptions'],
+  encounterDatetime: ['id', 'label', 'type', 'questionOptions', 'datePickerFormat'],
+  encounterLocation: ['id', 'label', 'type', 'questionOptions'],
+  encounterProvider: ['id', 'label', 'type', 'questionOptions'],
+  encounterRole: ['id', 'label', 'type', 'questionOptions'],
+  obs: ['id', 'label', 'type', 'questionOptions'],
+  obsGroup: ['id', 'label', 'type', 'questionOptions', 'questions'],
+  patientIdentifier: ['id', 'label', 'type', 'questionOptions'],
+  testOrder: ['id', 'label', 'type', 'questionOptions'],
+  programState: ['id', 'label', 'type', 'questionOptions'],
+};
+
+/**
+ * Mapping of allowed keys for the nested questionOptions object per question type.
+ */
+const allowedQuestionOptionsMapping: Record<string, string[]> = {
+  control: ['rendering', 'minLength', 'maxLength'],
+  encounterDatetime: ['rendering'],
+  encounterLocation: ['rendering'],
+  encounterProvider: ['rendering'],
+  encounterRole: ['rendering', 'isSearchable'],
+  obs: ['rendering', 'concept'],
+  obsGroup: ['rendering', 'concept'],
+  patientIdentifier: ['rendering', 'identifierType', 'minLength', 'maxLength'],
+  testOrder: ['rendering'],
+  programState: ['rendering', 'programUuid', 'workflowUuid', 'answers'],
+};
+
+/**
+ * Cleans the given questionOptions object by retaining only allowed keys for the new type.
+ */
+function cleanQuestionOptionsForType(options: any, newType: string): any {
+  const allowedOpts = allowedQuestionOptionsMapping[newType] || [];
+  const cleanedOpts = Object.fromEntries(Object.entries(options).filter(([optKey]) => allowedOpts.includes(optKey)));
+  // Ensure required property 'rendering' exists if present in the original options.
+  if (!('rendering' in cleanedOpts) && options.rendering) {
+    cleanedOpts.rendering = options.rendering;
+  }
+  return cleanedOpts;
+}
+
+/**
+ * Cleans the given form field by retaining only allowed top‑level properties for the new type.
+ * Also cleans nested questionOptions using the nested mapping.
+ */
+function cleanFormFieldForType(field: FormField, newType: string): FormField {
+  const allowedKeys = allowedPropertiesMapping[newType] || [];
+  const cleaned: Partial<FormField> = {} as Partial<FormField>;
+
+  // Copy only allowed top‑level properties.
+  (allowedKeys as (keyof FormField)[]).forEach((key) => {
+    if (key in field) {
+      (cleaned as any)[key] = field[key];
+    }
+  });
+
+  // If questionOptions is allowed and exists, clean it using the nested mapping.
+  if (cleaned.questionOptions && typeof cleaned.questionOptions === 'object') {
+    cleaned.questionOptions = cleanQuestionOptionsForType(cleaned.questionOptions, newType);
+  }
+
+  return { ...cleaned, type: newType } as FormField;
+}
+
 const QuestionModalContent: React.FC<QuestionModalProps> = ({
   formField: formFieldProp,
   closeModal,
@@ -40,6 +109,14 @@ const QuestionModalContent: React.FC<QuestionModalProps> = ({
 }) => {
   const { t } = useTranslation();
   const { formField, setFormField } = useFormField();
+  useEffect(() => {
+    if (formField && formField.type) {
+      const cleaned = cleanFormFieldForType(formField, formField.type);
+      if (JSON.stringify(cleaned) !== JSON.stringify(formField)) {
+        setFormField(cleaned);
+      }
+    }
+  }, [formField?.type, formField, setFormField]);
 
   const checkIfQuestionIdExists = useCallback(
     (idToTest: string): boolean => {
@@ -47,9 +124,7 @@ const QuestionModalContent: React.FC<QuestionModalProps> = ({
       const nestedIds = schema?.pages?.map((page) => {
         return page?.sections?.map((section) => {
           return section?.questions?.map((question) => {
-            question.questions?.map((nestedQuestion) => {
-              return nestedQuestion.id;
-            });
+            question.questions?.map((nestedQuestion) => nestedQuestion.id);
             return question.id;
           });
         });
@@ -64,7 +139,7 @@ const QuestionModalContent: React.FC<QuestionModalProps> = ({
     const emptyQuestion: FormField = {
       type: '',
       questionOptions: undefined,
-      id: '',
+      id: '', // new question without an id initially
     };
     setFormField((prevFormField) => ({
       ...prevFormField,
@@ -72,11 +147,12 @@ const QuestionModalContent: React.FC<QuestionModalProps> = ({
     }));
   }, [setFormField]);
 
+  // Updated deletion: filter out the question by its id.
   const deleteObsGroupQuestion = useCallback(
-    (index: number) => {
+    (id: string) => {
       setFormField((prevFormField) => ({
         ...prevFormField,
-        questions: prevFormField.questions?.filter((_, i) => i !== index) || [],
+        questions: prevFormField.questions?.filter((q) => q.id !== id) || [],
       }));
     },
     [setFormField],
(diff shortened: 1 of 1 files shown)


Bharath-K-Shetty:
@NethmiRodrigo review required..!

src/components/interactive-builder/modals/question/question.modal.tsx
```diff
@@ -163,10 +241,10 @@ const QuestionModalContent: React.FC<QuestionModalProps> = ({
                         <Question checkIfQuestionIdExists={checkIfQuestionIdExists} />
                         <Button
                           kind="danger"
-                          onClick={() => deleteObsGroupQuestion(index)}
+                          onClick={() => deleteObsGroupQuestion(question.id)}
                           className={styles.deleteObsGroupQuestionButton}
                         >
-                          {t('deleteQuestion', 'Delete question')}
+                          {t('deleteQuestion', 'Delete')}
```
harshthakkr, line 247: This change won’t affect the UI since the translation string needs to be updated first. The correct code should be: `{t('delete', 'Delete')}`.

harshthakkr (changes requested):

src/components/interactive-builder/modals/question/question.modal.tsx
Bharath-K-Shetty: Hi @harshthakkr,
I appreciate your suggestion. However, I’d like to keep the label as "Delete question" because it clearly indicates the action is specific to deleting a question. Since our translation file already includes the key "deleteQuestion" mapped to "Delete question", this provides the necessary context for users.

harshthakkr: No problem, looks fine!

NethmiRodrigo (changes requested):
Thanks @Bharath-K-Shetty! A couple of suggestions:

src/components/interactive-builder/modals/question/question.modal.tsx
```diff
@@ -81,11 +158,12 @@ const QuestionModalContent: React.FC<QuestionModalProps> = ({
...
   }, [setFormField]);
 
+  // Updated deletion: filter out the question by its id.
   const deleteObsGroupQuestion = useCallback(
-    (index: number) => {
+    (id: string) => {
       setFormField((prevFormField) => ({
         ...prevFormField,
-        questions: prevFormField.questions?.filter((_, i) => i !== index) || [],
+        questions: prevFormField.questions?.filter((q) => q.id !== id) || [],
```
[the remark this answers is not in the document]
Bharath-K-Shetty: i don't know how it changed maybe accidental..!i will revert it..!

Bharath-K-Shetty:
@NethmiRodrigo updated the pr..!

NethmiRodrigo (changes requested):
@Bharath-K-Shetty left a few comments. Please add unit tests to cover this scenario

src/components/interactive-builder/modals/question/question.modal.tsx
```diff
@@ -29,6 +29,86 @@ interface QuestionModalProps {
...
+  encounterDatetime: ['rendering'],
+  encounterLocation: ['rendering'],
+  encounterProvider: ['rendering'],
+  encounterRole: ['rendering', 'isSearchable'],
+  obs: ['rendering', 'concept', 'answers'],
+  obsGroup: ['rendering', 'concept'],
+  patientIdentifier: ['rendering', 'identifierType', 'minLength', 'maxLength'],
+  testOrder: ['rendering'],
+  programState: ['rendering', 'programUuid', 'workflowUuid', 'answers'],
+};
```
NethmiRodrigo, line 77: Lets move these into the constants file

src/components/interactive-builder/modals/question/question.modal.tsx
```diff
@@ -29,6 +29,86 @@ interface QuestionModalProps {
...
+
+  // If questionOptions is allowed and exists, clean it using the nested mapping.
+  if (cleaned.questionOptions && typeof cleaned.questionOptions === 'object') {
+    cleaned.questionOptions = cleanQuestionOptionsForType(cleaned.questionOptions, newType);
+  }
+
+  cleaned.type = newType;
+  return cleaned as FormField;
+}
+
```
NethmiRodrigo, line 111: I don't see a reason to have these functions defined here only to be exported to be used in a different component. It would be better if we move it to the function we use, or move these to a separate utils file

src/components/interactive-builder/modals/question/question-form/question/question.component.tsx
```diff
@@ -59,11 +60,14 @@ const Question: React.FC<QuestionProps> = ({ checkIfQuestionIdExists }) => {
...
       <Select
         value={formField?.type ?? 'control'}
-        onChange={(event: React.ChangeEvent<HTMLSelectElement>) =>
-          setFormField({ ...formField, type: event.target.value })
-        }
+        onChange={(event) => {
+          const newType = event.target.value;
+          const cleaned = cleanFormFieldForType({ ...formField, type: newType }, newType);
+          setFormField(cleaned);
+        }}
```
NethmiRodrigo, line 70: Its better to not do this because this would create a new function for every render (see our docs on guidance for performance - https://o3-docs.openmrs.org/docs/coding-conventions/performance). It would be better if we do 
```
const handleQuestionTypeChange = useCallback(() => 
 {
    // whatever the onchange handles
  }, []);
```
and then,
```suggestion
        onChange={handleQuestionTypeChange}
```

src/components/interactive-builder/modals/question/question-form/question/question.test.tsx
```diff
@@ -0,0 +1,45 @@
...
+        concept: '12345',
+        answers: [
+          { concept: '111', label: 'Yes' },
+          { concept: '222', label: 'No' },
+        ],
+        irrelevantOption: 'remove me',
+      },
+      required: true,
+      extraProp: 'should be removed',
+    } as any;
```
NethmiRodrigo, line 21: Why is it cast to as `any`?

NethmiRodrigo (changes requested):

src/components/interactive-builder/modals/question/question-form/question/question.test.tsx
Bharath-K-Shetty: @NethmiRodrigo  casting the test object as any because we intentionally add properties (like extraProp) that are not part of the official FormField interface. This lets us simulate a scenario where extra, irrelevant properties exist so that our cleaning function can be tested properly. Without casting to any, TypeScript would raise errors for these extra properties.

src/components/interactive-builder/modals/question/question-form/question/question.component.tsx
Bharath-K-Shetty: Ok understood..!

src/components/interactive-builder/modals/question/question-form/question/question.test.tsx
NethmiRodrigo: I understand that, but instead of casting to any to add properties that doesn't exist, why not leave it the type as is and instead make the test more realistic. i.e., in real life, the type wouldn't change to any, instead properties that exist would change. For example, changing the question type from `obs` to `control` and renderingType to `markdown` should remove the `label` attribute. Having tests like that is much better

src/components/interactive-builder/modals/question/question-utils.ts
```diff
@@ -0,0 +1,35 @@
+import type { FormField } from '@openmrs/esm-form-engine-lib';
+import { allowedPropertiesMapping, allowedQuestionOptionsMapping } from '../../../../constants';
```
NethmiRodrigo, line 2:
```suggestion
import { allowedPropertiesMapping, allowedQuestionOptionsMapping } from '@constants';
```

src/components/interactive-builder/modals/question/question.modal.tsx
```diff
@@ -45,10 +45,6 @@ const QuestionModalContent: React.FC<QuestionModalProps> = ({
 }) => {
   const { t } = useTranslation();
   const { formField, setFormField } = useFormField();
-
-  /**
-   * NOTE - this does not support nested obsGroup questions
-   */
   const checkIfQuestionIdExists = useCallback(
```
NethmiRodrigo, line 48: Was removing this accidental?

translations/en.json
NethmiRodrigo, line 1: I don't think these translations are a part of this diff. Were these accidental?

src/components/interactive-builder/modals/question/question-form/question/question.component.tsx
```diff
@@ -31,6 +32,15 @@ const Question: React.FC<QuestionProps> = ({ checkIfQuestionIdExists }) => {
     return checkIfQuestionIdExists(formField.id);
   }, [formField.id, checkIfQuestionIdExists]);
 
+  const handleQuestionTypeChange = useCallback(
+    (event: React.ChangeEvent<HTMLSelectElement>) => {
+      const newType = event.target.value;
+      const cleaned = cleanFormFieldForType({ ...formField, type: newType }, newType);
+      setFormField(cleaned);
+    },
+    [formField, setFormField],
```
NethmiRodrigo, line 41: Having formField and setFormField as dependencies aren't good because setFormField updated the formField. Check the other functions to see a better way to use setFormField - https://github.com/openmrs/openmrs-esm-form-builder/blob/eff9e72db3964140f0c1068d9086397a45236520/src/components/interactive-builder/modals/question/question-form/question-types/inputs/obs/obs-type-question.component.tsx#L32

NethmiRodrigo (changes requested):

Bharath-K-Shetty:
Hey @NethmiRodrigo i think for cleanup of rendering Type there is a separate ticket focusing on that can you please clarify once.

NethmiRodrigo:
> Hey @NethmiRodrigo i think for cleanup of rendering Type there is a separate ticket focusing on that can you please clarify once.Ticket:https://openmrs.atlassian.net/browse/O3-4319

Yeah, I thought we could handle it separately, but now seeing it in action, I don't think we can do them seperately

src/constants.ts
```diff
@@ -54,3 +54,50 @@ export const renderTypeOptions: Record<Exclude<QuestionType, 'obs'>, Array<Rende
...
+/**
+   Type-specific properties for each question type.
+ */
+export const typeSpecificProperties: Record<string, Array<keyof FormField>> = {
+  control: [],
+  encounterDatetime: ['datePickerFormat'],
+  encounterLocation: [],
+  encounterProvider: [],
+  encounterRole: [],
+  obs: ['required'],
```
NethmiRodrigo, line 72: 'required` is a common property

src/constants.ts
```diff
@@ -54,3 +54,50 @@ export const renderTypeOptions: Record<Exclude<QuestionType, 'obs'>, Array<Rende
...
+    const mergedProps = new Set<string>([...requiredProperties, ...props]);
+    return [type, Array.from(mergedProps)];
+  }),
+);
+
+/**
+   Mapping of allowed keys for the nested questionOptions object per question type.
+ */
+export const allowedQuestionOptionsMapping: Record<string, string[]> = {
+  control: ['rendering', 'minLength', 'maxLength'],
```
NethmiRodrigo, line 93: 'minLength', 'maxLength' are properties relevant to `text` rendering type and not specific to the `control` question type

src/constants.ts
```diff
@@ -54,3 +54,50 @@ export const renderTypeOptions: Record<Exclude<QuestionType, 'obs'>, Array<Rende
...
+  encounterDatetime: ['rendering'],
+  encounterLocation: ['rendering'],
+  encounterProvider: ['rendering'],
+  encounterRole: ['rendering', 'isSearchable'],
```
NethmiRodrigo, line 97: Similarly, `isSearchable` is for `ui-select-extended` rendering type

NethmiRodrigo (commented):
@Bharath-K-Shetty you'd have to reconsider the properties you added for questionType since some of them are for renderingType

NethmiRodrigo:
@Bharath-K-Shetty  So what I do to find the correct properties is, I go through the FormField type definition - https://github.com/openmrs/openmrs-esm-form-engine-lib/blob/d28f37f92cff6c6e97b3339113d921e880400855/src/types/schema.ts#L60 because most often than not they are pretty self explanatory. Similarly, - https://openmrs.atlassian.net/wiki/spaces/docs/pages/68681754/Field+Types+Reference for `renderingType`. Most of what you have put for `questionType` you'd see is actually what should be under `renderingType`

NethmiRodrigo (commented):
@Bharath-K-Shetty I've left a few comments, but the more I think about this, the more this approach seems to be unscalable, especially because I was thinking of all the possible combinations of `questionType` to `renderingType` we'd have to consider. Seems like the best way forward is to rework this to:
1. When changing the question type or rendering type, we just remove all the properties except id, label and required
2. Show a info tooltip next to the question type and rendering type that on hover would show a warning that changing the type would remove properties.
Over time, we'll probably evolve this into a better solution, but I think this proposed approach would be a much better and less complex task (and easier to test). 
Sorry about the iterative changes, this is a ticket that I hadn't thought about in such depth until now

Bharath-K-Shetty:
Hey @NethmiRodrigo should i close it and start it again because it has became a mess now..!

PR #426, opened 2025-03-30
(fix) O3-4581: Reset the rendering type when the question type changes

yoursanonymous:
Added logic to remove rendering type value if the previously selected rendering type value is not present for the new question type

## Requirements

## Summary
Rendering type was same even after changing the Question type and in UI it selects the 1st dropdown value if previous selected rendering type is not available

## Related Issue
https://openmrs.atlassian.net/browse/O3-4581


Pratibha0280:
Hey @yoursanonymous , I noticed that the code directly modifies the formField object using the delete operator.
Instead, I would suggest to create a new object without the questionOptions property when needed. This way, the state updates cleanly and predictably.

Pratibha0280 (commented):
Hey @yoursanonymous , I noticed that the code directly mutates the formField object using the delete operator.
Instead, I would suggest to create a new object without the questionOptions property when needed. This way, the state updates cleanly and predictably.

src/components/interactive-builder/modals/question/question-form/question/question.component.tsx
```diff
@@ -32,6 +32,18 @@ const Question: React.FC<QuestionProps> = ({ checkIfQuestionIdExists }) => {
...
+    setFormField((formField)=>{
+      if(formField?.questionOptions?.rendering && questionTypeNew!=='obs' &&  
+      questionTypes.includes(questionTypeNew as keyof typeof renderTypeOptions) &&
+      !renderTypeOptions[questionTypeNew].includes(formField.questionOptions.rendering as  RenderType)){
+        formField.questionOptions.rendering = null;
+        return { ...formField, type: questionTypeNew};
+      }
+      return { ...formField, type: questionTypeNew };
+    }); 
+  }
```
NethmiRodrigo, line 45: A couple of things, prefer to not use inline function calls - https://o3-docs.openmrs.org/docs/coding-conventions/performance, and since a variable of `formField` is already there, it can be confusing. And the way this whole if function is written is very hard to comprehend, its preferred to break it down

NethmiRodrigo: Maybe something like this (do not directly use this suggestion, I haven't looked much into this suggested code), the gist is that the conditions should be broken down into logical conditions that make sense when considered together:
```suggestion
  const handleQuestionTypeChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
    const newQuestionType = event.target.value;
    setFormField((formField)=>{
     if (!currentField.questionOptions || !currentField.questionOptions?.rendering) {
        return { ...currentField, type: newQuestionType };
      }

      const currentRendering = currentField.questionOptions.rendering;

      if (!currentRendering || newQuestionType === 'obs') {
        return { ...currentField, type: newQuestionType };
      }

      const shouldResetRendering =
        newQuestionType in questionTypes &&
        !renderTypeOptions[newQuestionType].includes(currentRendering as RenderType);

      if (shouldResetRendering) {
        return {
          ...currentField,
          type: newQuestionType,
          questionOptions: {
            ...currentField.questionOptions,
            rendering: null,
          },
        };
      }

      return { ...currentField, type: newQuestionType };
    }); 
  }
```

NethmiRodrigo (changes requested):

src/components/interactive-builder/modals/question/question-form/question/question.component.tsx
```diff
@@ -45,6 +45,35 @@ const Question: React.FC<QuestionProps> = ({ checkIfQuestionIdExists }) => {
...
+  
+  const handleQuestionTypeChange = useCallback(
+    (event: React.ChangeEvent<HTMLSelectElement>) => {
+      const newQuestionType = event.target.value;
+      setFormField((prevFormField) => {
+        const hasPreviousRenderingType = prevFormField?.questionOptions?.rendering;
+        if (hasPreviousRenderingType) {
+          const isQuestionTypeObs = newQuestionType === 'obs' ? true : false;
+          if (!isQuestionTypeObs) {
+            const isNewQuestionTypeHasPrevRenderingType =
```
NethmiRodrigo, line 57: Nit:
```suggestion
            const isRenderingTypeValidForQuestionType =
```

src/components/interactive-builder/modals/question/question-form/question/question.component.tsx
```diff
@@ -45,6 +45,35 @@ const Question: React.FC<QuestionProps> = ({ checkIfQuestionIdExists }) => {
...
+              questionTypes.includes(newQuestionType as keyof typeof renderTypeOptions) &&
+              renderTypeOptions[newQuestionType].includes(prevFormField.questionOptions.rendering as RenderType);
+            if (!isNewQuestionTypeHasPrevRenderingType) {
```
NethmiRodrigo, line 60:
```suggestion
            if (!isRenderingTypeValidForQuestionType) {
```

NethmiRodrigo (approved):

yoursanonymous:
Hi @NethmiRodrigo , Thankyou for approving my PR, 
Is it alright to merge it now?

NethmiRodrigo:
Thanks @yoursanonymous!

PR #403, opened 2025-02-19
(fix) O3-4314: Add Drag-and-Drop Reordering for Selected Answers with Portal Integration

Bharath-K-Shetty:
## Requirements

- [ ] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-docs.openmrs.org/docs/frontend-modules/contributing.en-US#contributing-guidelines) label. See existing PR titles for inspiration.
- [ ] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide) and [design documentation](https://om.rs/o3ui).
- [ ] My work includes tests or is validated by existing tests.

## Summary

This PR introduces drag-and-drop functionality to the Select Answers component in the form builder, enabling users to rearrange their selected answers to match their preferred display order.

Key changes include:
- Integration of react-beautiful-dnd components (DragDropContext, Droppable, Draggable) to manage drag events.
- Implementation of the handleDragEnd function to update the answer order based on user interaction.
- Use of a DraggablePortal to render the dragged element at the document body level, resolving offset issues caused by CSS transforms in modal containers.
- Minor updates to state management to ensure the new order is immediately reflected.

## Screenshots

https://github.com/user-attachments/assets/31d35898-ae3d-47e1-a6d2-b4fd203df4bb

## Related Issue

https://openmrs.atlassian.net/browse/O3-3914

PR #445, opened 2025-04-08
(feat) O3-4171 : Validate concept answers and show error for non-existing answers

Bharath-K-Shetty:
## Requirements

## Summary

-  Checks each answer concept against the original concept's answer list
-  Performs a concept lookup only for non-original answers
-  Shows an inline notification when one or more answers are invalid.

## Screenshots

https://github.com/user-attachments/assets/79d9856c-f182-49f5-9594-f925b764f71e

## Related Issue

https://openmrs.atlassian.net/browse/O3-4171


Bharath-K-Shetty:
Hey @NethmiRodrigo should i stick with this endpoint `${restBaseUrl}/concept?q=${conceptId}&v=full` for additional concept answers validation or current end point is good? 
Also, I'm planning to extend the validation to check for invalid concept answer names as well. I haven’t implemented that part yet since there are a few existing issues with editing concept answer names that I want to avoid interfering with for now.

src/components/interactive-builder/modals/question/question-form/rendering-types/inputs/select/select-answers.component.tsx
```diff
@@ -124,6 +126,46 @@ const SelectAnswers: React.FC = () => {
...
+          } catch (error) {
+            allValid = false;
+            break;
+          }
+        }
+      }
+      setInvalidAnswerFound(!allValid);
+    };
+
+    validateAnswers();
```
NethmiRodrigo, line 166: Since you are calling this function inside useEffect, the function should be defined outside the `useEffect` hook.

src/components/interactive-builder/modals/question/question-form/rendering-types/inputs/select/select-answers.component.tsx
```diff
@@ -124,6 +126,46 @@ const SelectAnswers: React.FC = () => {
...
+
+      let allValid = true;
+      const lookupCache = new Map<string, boolean>();
+      const uniqueAnswers = Array.from(new Map(answerItems.map((a) => [a.id, a])).values());
+
+      for (const answer of uniqueAnswers) {
+        if (originalAnswerIds.has(answer.id)) {
+          continue;
+        } else {
+          const url = `${restBaseUrl}/concept/${answer.id}?v=full`;
```
NethmiRodrigo, line 147: Please use the `useConceptId` hook. 
For future reference, we don't usually do API calls within the component like this. We make hooks  - https://o3-docs.openmrs.org/docs/coding-conventions/data-fetching.

src/components/interactive-builder/modals/question/question-form/rendering-types/inputs/select/select-answers.component.tsx
```diff
@@ -152,6 +194,17 @@ const SelectAnswers: React.FC = () => {
...
+      {/* Display an inline notification if any answer fails validation */}
+      {invalidAnswerFound && (
+        <InlineNotification
+          kind="error"
+          lowContrast
+          className={styles.error}
+          title={t('invalidAnswerConcept', 'Invalid Answer Concept Detected')}
+          subtitle={t('answerConceptValidation', 'One or more selected answer concepts do not exist in the system. ')}
+        />
+      )}
```
NethmiRodrigo, line 206: This is good, maybe we can show a not found icon within the blue tag of the answer that is invalid. Otherwise we'd have to guess which ones are invalid.

NethmiRodrigo (changes requested):

src/components/interactive-builder/modals/question/question-form/rendering-types/inputs/select/select-answers.component.tsx
Bharath-K-Shetty: @NethmiRodrigo But we can't directly use the hook inside the loop right??

Bharath-K-Shetty:
Screenshot 2025-04-15 175423
Added Warning icon for invalid concept answers..!

src/components/interactive-builder/modals/question/question-form/rendering-types/inputs/select/select-answers.component.tsx
```diff
@@ -130,6 +132,43 @@ const SelectAnswers: React.FC = () => {
...
+
+    const invalidIds: string[] = [];
+    const uniqueAnswers = Array.from(new Map(answerItems.map((a) => [a.id, a])).values());
+
+    for (const answer of uniqueAnswers) {
+      if (originalAnswersMap.has(answer.id)) {
+        const expectedName = originalAnswersMap.get(answer.id);
+        if (answer.text !== expectedName) {
+          invalidIds.push(answer.id);
+        }
```
NethmiRodrigo, line 151: Hey @Bharath-K-Shetty, does this mean that if the label of the answer is not the same as the label of the original answer from the concept, it will be flagged as invalid?

NethmiRodrigo (approved):
Any progress on fixing the test?

Bharath-K-Shetty:
Hey @NethmiRodrigo & @denniskigen i think i got the solution instead of explicitly adding the icon.In `<Tag>` we have a prop called decorator where we can add a warning icon..!

denniskigen:
Isn't that experimental though? It looks like you can use the `renderIcon` prop to achieve the same goal.

Bharath-K-Shetty:
Hey @denniskigen Yeah i understand that it is experimental.But if i use `renderIcon` prop instead of `decorator` then the icon will be rendered first before the child..!

NethmiRodrigo:
> But if i use `renderIcon` prop instead of `decorator` then the icon will be rendered first before the child..!

@Bharath-K-Shetty that's actually fine

Bharath-K-Shetty:
@NethmiRodrigo fine then let me push that changes..!

src/components/interactive-builder/modals/question/question-form/rendering-types/inputs/select/select-answers.component.tsx
Bharath-K-Shetty: Yes @NethmiRodrigo

NethmiRodrigo: Is there a reason behind that? Seems like if the concept UUID is correct, that is sufficient, since we fetch using the UUID only.

Bharath-K-Shetty: Here i added validation for names too because in the schema we can edit the concept answer to a name which isn't present in the system(even with the valid UUID) which isn't incorrect right? Also it is already been used for validating the concept answers so i thought it would be better if validation of names also be there & related to the fetching it ,it also returns the labels of the valid  concepts right..?

NethmiRodrigo: How I was thinking about it is, if changing the label of a concept doesn't cause a failure when saving the form, then the form builder shouldn't flag an error saying concept not found. 
> Also, since label validation is already used for validating concepts

Mind pointing me to where we're doing that?

> Regarding the fetching part—it also returns the labels of the valid concepts, which we can use to identify and validate any invalid ones.

Yes, but we aren't using the label to fetch concepts, so it feels wrong to invalidate saying `concept not found`, when it can be found.

Bharath-K-Shetty: > Can you test this out and see? Change just the label of an answer in the form, and submit the form in the patient chart for a patient, and see if it causes an error? If it doesn't then we shouldn't show any error for it.

but in the patient chart it is not showing error for both invalid id and label of concept answer..?

NethmiRodrigo: > but in the patient chart it is not showing error for both invalid id and label of concept answer..?(It is showing error if main selected concept answer has invalid id)

I don't understand. What do you mean main selected concept answer? What is a concept answer, and what is a main selected concept answer?

NethmiRodrigo: If on patient chart you didn't get an error saving a form that had a answer whose label you you altered, then yes you should remove the validation done on labels.

Bharath-K-Shetty: > I don't understand. What do you mean main selected concept answer? What is a concept answer, and what is a main selected concept answer?
What I meant here is it is not showing error for both invalid id and label of a concept answer.(But it is showing error for the main backing concept which we selected if it has invalid id)

Bharath-K-Shetty:
Updated @NethmiRodrigo ..!

src/components/interactive-builder/modals/question/question-form/rendering-types/inputs/select/select-answers.component.tsx
```diff
@@ -1,11 +1,12 @@
...
+import { Tag, MultiSelect, Stack, InlineNotification } from '@carbon/react';
+import { WarningAltFilled } from '@carbon/react/icons';
 import { useTranslation } from 'react-i18next';
 import ConceptSearch from '../../../common/concept-search/concept-search.component';
 import { useFormField } from '../../../../form-field-context';
+import { fetchConceptById } from '@resources/concept.resource';
 import type { Concept } from '@types';
 import styles from './select-answers.scss';
-
 interface AnswerItem {
```
NethmiRodrigo, line 10: Nit: I think you've accidentally deleted a new line here
```suggestion

interface AnswerItem {
```

NethmiRodrigo (commented):
@Bharath-K-Shetty is the plan to merge this in and then work on updating the UI for the answer items in #403?

Bharath-K-Shetty:
> @Bharath-K-Shetty is the plan to merge this in and then work on updating the UI for the answer items in #403?

yes @NethmiRodrigo ..!

PR #446, opened 2025-04-08
O3-4613 Choosing a default radio button "type of date picker to show" radio group

yoursanonymous:
…roup button

## Requirements

### DIFF
--- a/src/components/interactive-builder/modals/question/question-form/rendering-types/inputs/date/date.component.tsx
+++ b/src/components/interactive-builder/modals/question/question-form/rendering-types/inputs/date/date.component.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback } from 'react';
+import React, { useCallback, useEffect } from 'react';
 import { RadioButtonGroup, RadioButton } from '@carbon/react';
 import { useTranslation } from 'react-i18next';
 import { useFormField } from '../../../../form-field-context';
@@ -14,6 +14,18 @@ const Date: React.FC = () => {
     time: [{ value: 'timer', label: t('timerOnly', 'Timer only'), defaultChecked: false }],
   };
 
+  const initializeDefaultSelection=useCallback(()=>{
+    if(!formField.datePickerFormat){
+      setFormField((prevField)=>({
+        ...prevField,
+        datePickerFormat:'both'
+      }));
+    }
+  },[formField.datePickerFormat,setFormField]);
+  useEffect(()=>{
+    initializeDefaultSelection();
+  },[initializeDefaultSelection])
+  
   const handleDatePickerTypeChange = useCallback(
     (type: DatePickerTypeOption) => {
       setFormField((prevField) => ({


atulyadav745:
Thanks @yoursanonymous, great catch! Really appreciate you pointing this out.

NethmiRodrigo (changes requested):
I don't think its really necessary to default pick a datepicker type for this, but if we are doing it, its much better to set it here - https://github.com/openmrs/openmrs-esm-form-builder/blob/b5221ee782c419db7cbb1020237b554f05302ac5/src/components/interactive-builder/modals/question/question-form/question/question.component.tsx#L90 rather than use an useEffect. We should avoid the usage of useEffects wherever we can - https://react.dev/learn/you-might-not-need-an-effect

yoursanonymous:
Hi @NethmiRodrigo , Sorry just realized by looking at a similar 'ui-select-extended' component that using 'defaultSelected' property of RadioButtonGroup would be a better way to set the default value for Date picker type,  made the required changes as well. 
Thank you for reviewing my PR, could you please review it again.

src/components/interactive-builder/modals/question/question-form/rendering-types/inputs/date/date.component.tsx
```diff
@@ -25,7 +25,7 @@ const Date: React.FC = () => {
   );
 
   return (
-    <RadioButtonGroup name="datePickerType" legendText={t('datePickerType', 'The type of date picker to show ')}>
+    <RadioButtonGroup name="datePickerType" defaultSelected="both" legendText={t('datePickerType', 'The type of date picker to show ')}>
```
NethmiRodrigo, line 28: >  using 'defaultSelected' property of RadioButtonGroup would be a better way to set the default value for Date picker type, made the required changes as well.

You can use `defaultSelected` but that would mean that the user gets the impression that when they save the defaultSelected value would be persisted, but that isn't the case here. To fix that, referring back to my first suggestion, you'd have to set the `datePickerFormat` in the `formField`

NethmiRodrigo (changes requested):

yoursanonymous:
Hi @NethmiRodrigo, after trying out multiple scenarios the issue was found to be pretty big one in which when we change the rendering type, the subsequent subproperties of previous rendering type was not removed from the Form json and the default value selected for various fields in interactive builder was also not added automatically by default to json object.

yoursanonymous:
https://github.com/user-attachments/assets/ee0353ba-7d02-49c0-b057-281e66a38778


https://github.com/user-attachments/assets/162f71fd-bd80-43b4-9efd-64aab0a554d1

yoursanonymous:
Hi @NethmiRodrigo ,I have also fixed this error in question component


https://github.com/user-attachments/assets/516f30a7-3569-4365-bb43-162dcad1013c


https://github.com/user-attachments/assets/4c20acde-7142-403a-8c05-c6adf7873109

yoursanonymous:
Removing unnecesary fields during questionType change.


https://github.com/user-attachments/assets/95950590-5603-4abe-b0a8-d8fa4dd424ce



https://github.com/user-attachments/assets/e47bcf84-a7b2-4340-81ff-beabe68789d4

yoursanonymous:
Hi @NethmiRodrigo, awaiting your approval, Kindly review

NethmiRodrigo:
> Hi @NethmiRodrigo, after trying out multiple scenarios the issue was found to be pretty big one in which when we change the rendering type, the subsequent subproperties of previous rendering type was not removed from the Form json and the default value selected for various fields in interactive builder was also not added automatically by default to json object.

@yoursanonymous The issue you've pivoted to address is already being worked on in PR #391, which has several comments and insights on how to fix it properly.
You have two good options moving forward:
1. Revert this PR back to your original scope (addressing the initial ticket with the suggestions I provided in my first two reviews, which is sufficient to fix your initial problem, this current issue that you are addressing doesn't affect that)
2. Collaborate on PR #391 if you're particularly interested in solving that specific issue

For future reference, it's best practice to keep PRs focused on their original scope. If you want to work on a different issue, please open a new PR rather than changing direction mid-review. This wastes reviewer time and disrupts our development process. One PR = one focused purpose.
I'm happy to help get your original work merged if you decide to go back to option 1.

PR #279, opened 2024-05-25
feat(O3-3230): add rule builder to form builder

senthil-k8s:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-docs.openmrs.org/docs/frontend-modules/contributing.en-US#contributing-guidelines) label. See existing PR titles for inspiration.
- [x] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide) and [design documentation](https://om.rs/o3ui).
- [x] My work includes tests or is validated by existing tests.

## Summary
This PR integrates rule builder component into form builder, which provides an user-interface rule builder to build conditional logic for the forms.

## Screenshots
https://www.loom.com/share/3946df798a014697b595ea87140db7cc

## Related Issue
https://openmrs.atlassian.net/browse/O3-3230

## Other
Figma Design of rule builder:
Final UIUX of Rule builder

PR #450, opened 2025-04-09
(feat) O3-3230: Add rule builder to interactive builder

NethmiRodrigo:
## Requirements

## Summary

This PR integrates rule builder component into form builder, which provides an user-interface rule builder to build conditional logic for the forms in the form builder. Basically the PR #279, I've made a new PR because I worked off a new branch.

## Related Issue

https://issues.openmrs.org/browse/O3-3230


NethmiRodrigo:
I swear to god, this e2e test passes locally
Screenshot 2025-04-09 at 9 59 15 PM
Screenshot 2025-04-09 at 10 00 11 PM

src/components/rule-builder/conditional-logic.tsx
```diff
@@ -0,0 +1,52 @@
...
+  rules,
+  schema,
+  onSchemaChange,
+  handleAddLogic,
+}) => {
+  const rulesForQuestionIndex = rules?.findIndex((rule: FormRule) => rule?.question === question?.id);
+  const rulesForQuestion =
+    rulesForQuestionIndex !== -1 && rulesForQuestionIndex !== undefined
+      ? rules?.filter((rule: FormRule) => rule?.question === question?.id)
+      : [{ id: uuidv4(), isNewRule: false, question: question?.id }];
```
ibacher, line 31: This seems like it could be re-written to be much simpler:

```suggestion
const rulesForQuestion = rules?.filter((rule: FormRule) => rule?.question === question?.id) ??
      [{ id: uuidv4(), isNewRule: false, question: question?.id }];
```

src/components/rule-builder/conditional-logic.tsx
```diff
@@ -0,0 +1,52 @@
...
+  return (
+    <>
+      {rulesForQuestion.map((rule: FormRule) => (
+        <RuleBuilder
+          ruleId={rule.id}
+          key={question.id}
```
ibacher, line 37: Isn't `question.id` basically guaranteed to be non-unique? I.e., every value of `rulesForQuestion` should have the same `question.id`, right?

src/components/rule-builder/conditional-logic.tsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import type { Question, Schema } from '@types';
+import RuleBuilder, { type FormRule } from './rule-builder.component';
+import { v4 as uuidv4 } from 'uuid';
```
ibacher, line 4:
```suggestion
import { v4 as uuid } from 'uuid';
```

src/components/rule-builder/delete-conditional-logic.modal.tsx
```diff
@@ -0,0 +1,269 @@
...
+              validatorIndex - 1
+            ]
+          : undefined;
+      const existingValidatorIndex = validators.findIndex((validator) => validator === existingValidator);
+      if (existingValidatorIndex !== -1) {
+        validators?.splice(existingValidatorIndex, 1);
+      }
+      setRules((prevRules) => {
+        const updatedRules = prevRules.map((rule) => {
+          const updatedRule = { ...rule };
```
ibacher, line 76: I don't think this is necessary. You can safely modify `prevRules` without negative effects.

src/components/rule-builder/input-selection-box.component.tsx
```diff
@@ -0,0 +1,156 @@
...
+    const [isDropdownVisible, setIsDropdownVisible] = useState(false);
+    const [searchTermValue, setSearchTermValue] = useState<string>('');
+    const conditionalValue = useDebounce(inputValue, 500);
+    const selectedSearchValue = useDebounce(searchTermValue, 500);
+    const { t } = useTranslation();
+    const dropdownRef = useRef<HTMLUListElement>(null);
+
+    useEffect(() => {
+      if (searchTermValue.length) onChange(conditionalValue);
+      // eslint-disable-next-line react-hooks/exhaustive-deps
```
ibacher, line 37: It seems dangerous not to include `searchTermValue.length` here? Sometimes ignoring deps in `useEffect()` makes sense, but that's usually where there's another `useEffect()` hook that applies the changes in the cases this mises.

src/components/rule-builder/input-selection-box.component.tsx
```diff
@@ -0,0 +1,156 @@
+import React, { useEffect, useState, useRef, useCallback, useMemo } from 'react';
+import { TextInput, MultiSelect } from '@carbon/react';
+import { useDebounce } from '@openmrs/esm-framework';
+import { useTranslation } from 'react-i18next';
+import styles from './input-selection-box.scss';
+
+interface CustomComboBoxProps {
+  id: string;
+  items: Array<Record<string, string>>;
```
ibacher, line 9: Shouldn't the type here be a little more specific? I.e., we require a `label` property for this to work and it seems like we also expect a `concept` property... and ignore anything else.

src/components/rule-builder/input-selection-box.scss
```diff
@@ -0,0 +1,49 @@
...
+  top: 50%;
+  transform: translateY(-60%);
+  cursor: pointer;
+  border: none;
+  background: none;
+  font-size: 1rem;
+}
+.combobox .multiSelect {
+  position: relative;
+  bottom: 8px;// Adjust the z-index to a higher value to ensure it's on top
```
ibacher, line 48: 1. There are a bunch of rules here specified using `px`. All of these should be changed to `rem` measurements.
2. The comment here is, um... ???

src/components/rule-builder/rule-builder.component.tsx
```diff
@@ -0,0 +1,1746 @@
...
+    const [isAllowFutureDate, setIsAllowFutureDate] = useState<boolean>(
+      question?.validators?.some((validator) =>
+        validator.type === RenderingType.DATE && validator?.allowFutureDates === 'true' ? true : false,
+      ),
+    );
+    const [ruleDeleted, setRuleDeleted] = useState<boolean>(false);
+
+    const prevPageIndex = useRef(-1);
+    const prevSectionIndex = useRef(-1);
+    const prevQuestionIndex = useRef(-1);
```
ibacher, line 132: I'm a little concerned that these are really intended to be _states_ rather than _refs_.

src/components/rule-builder/rule-builder.component.tsx
```diff
@@ -0,0 +1,1746 @@
...
+      schema,
+      sectionIndex,
+    ]);
+
+    /**
+     * Build the schema for a "target" condition.
+     * Generates the condition expression based on the provided parameters.
+     */
+    const getSchemaCondition = (
+      condition: string,
```
ibacher, line 233: Shouldn't `condition` here have a stricter type, i.e., to correspond to the set of known conditions?

src/components/rule-builder/rule-builder.component.tsx
```diff
@@ -0,0 +1,1746 @@
...
+      [],
+    );
+
+    /**
+     * Retrieve the arguments required for calculation functions.
+     * Identifies and returns the necessary parameters based on the type of calculation.
+     */
+    const getArguments = useCallback((expression: string) => {
+      switch (expression) {
+        case 'BMI':
```
ibacher, line 357: In general, we seem to have a lot of these that switch on labels rather than some sort of language-independent key, which means that this makes it harder to (eventually) translate this part of the app.

ibacher (commented):
Thanks @NethmiRodrigo! I realize a bunch of these are nit-picky, but I think one or two of these really need to be addressed.

src/components/rule-builder/conditional-logic.tsx
NethmiRodrigo: Yeah, looks like it doesn't make sense for the key to be the `question.id` in that case, does it?

ibacher: Yeah. I think it's probably fine to use `rule.id` though? That should be unique-per-tree.

PR #486, opened 2025-04-27
(feat) O3-3596: Support for referencing reusable components from all forms

MohamedAliMefteh:
## Requirements

## Summary
This PR introduces the ability to reference components from any saved form during form building, expanding beyond only using saved "component forms".

### Specifically:

* Next to the "Add Section" button, an "Add Reference" button is added.

* Clicking "Add Reference" opens a modal:

  * A dropdown allows the user to select from all available forms (not just component forms).

  * Once a form is selected, another dropdown renders to allow the user to select a page from that form.

  * After selecting a page, a radio button group displays all the sections in the page.

  * After selecting a section, a list of checkboxes for the questions in that section appears (all checked by default).

  * Users can uncheck any questions they want to exclude.

* The "Save" button is enabled only when:

  * A form, page, section, and at least one question are selected.

* Upon saving:

  * The form schema is modified to include the new reference in referencedForms.

  * The selected section is added as a referenced section.

### Additionally:

* In the Interactive Builder, modifications were made to:

  * Prevent users from editing referenced sections directly (no adding/modifying questions inside referenced sections).

  * Allow users to delete a referenced section if needed.

This feature allows more flexibility when building forms by enabling reuse of any existing form content — not limited to specially-created component forms.

## Screenshots

https://github.com/user-attachments/assets/6227ee19-03ef-4e0a-aada-6855d63ec313

## Related Issue
https://openmrs.atlassian.net/browse/O3-3596


src/components/interactive-builder/modals/add-form-reference/add-form-reference.scss
```diff
@@ -0,0 +1,9 @@
+@use '@carbon/layout';
+
+.modalBody {
+  overflow-y: scroll;
+  min-height: 300px;
+}
+.form {
```
NethmiRodrigo, line 7:
```suggestion
}

.form {
```

src/components/interactive-builder/modals/add-form-reference/add-form-reference.modal.tsx
```diff
@@ -0,0 +1,241 @@
...
+  FormGroup,
+  RadioButtonGroup,
+  RadioButton,
+  Stack,
+} from '@carbon/react';
+import { type Form as FormType, type Schema, type Page, type Section, type Question } from '@types';
+import { useForms } from '@hooks/useForms';
+import { useClobdata } from '@hooks/useClobdata';
+import styles from './add-form-reference.scss';
+import { showSnackbar } from '@openmrs/esm-framework';
```
NethmiRodrigo, line 23: Please re-arrange the imports

src/components/interactive-builder/modals/add-form-reference/add-form-reference.modal.tsx
```diff
@@ -0,0 +1,241 @@
...
+            })
+          : (schema.referencedForms = [
+              {
+                formName: selectedForm.name,
+                alias: selectedForm.name,
+              },
+            ]);
+      }
+      schema.pages[pageIndex].sections.push({
+        label: `${selectedSection.label} (This section is added from ${selectedForm.name})`,
```
NethmiRodrigo, line 102: Its not necessary to add this text to the label, especially since we don't let them edit this. If its a concern that then the user would not know that its a referenced component from the overview in the interactive builder until they expand the accordion, we can add a tag next to the label saying `Referenced Component`.
```suggestion
        label: selectedSection.label,
```

src/components/interactive-builder/modals/add-form-reference/add-form-reference.modal.tsx
```diff
@@ -0,0 +1,241 @@
...
+  return (
+    <>
+      <ModalHeader closeModal={closeModal} title={t('addFormComponent', 'Add Form Component')} />
+      <Form className={styles.form}>
+        <ModalBody className={styles.modalBody}>
+          <Stack gap={4}>
+            {isLoading ? (
+              <InlineLoading description={t('loading', 'Loading...')} />
+            ) : error ? (
+              <InlineNotification>{t('errorLoadingForms', 'Error loading Forms')}</InlineNotification>
```
NethmiRodrigo, line 139:
```suggestion
              <InlineNotification>{t('errorLoadingForms', 'Error loading forms')}</InlineNotification>
```

src/components/interactive-builder/modals/add-form-reference/add-form-reference.modal.tsx
```diff
@@ -0,0 +1,241 @@
...
+            ) : forms.length === 0 ? (
+              <InlineNotification>{t('noFormsAvailable', 'No Forms Available')}</InlineNotification>
```
NethmiRodrigo, line 141:
```suggestion
              <InlineNotification>{t('noFormsAvailable', 'No forms available')}</InlineNotification>
```

src/components/interactive-builder/modals/add-form-reference/add-form-reference.modal.tsx
```diff
@@ -0,0 +1,241 @@
...
+                  }))}
+                  itemToString={(item: FormType) => (item ? item.name : '')}
+                  onChange={selectForm}
+                />
+              </FormGroup>
+            )}
+            {isLoadingClobdata ? (
+              <InlineLoading description={t('loading', 'Loading...')} />
+            ) : clobdataError ? (
+              <InlineNotification>{t('errorLoadingForm', 'Error loading Form')}</InlineNotification>
```
NethmiRodrigo, line 161:
```suggestion
              <InlineNotification>{t('errorLoadingForm', 'Error loading form')}</InlineNotification>
```

src/components/interactive-builder/modals/add-form-reference/add-form-reference.modal.tsx
```diff
@@ -0,0 +1,241 @@
...
+            ) : error ? (
+              <InlineNotification>{t('errorLoadingForms', 'Error loading Forms')}</InlineNotification>
+            ) : forms.length === 0 ? (
+              <InlineNotification>{t('noFormsAvailable', 'No Forms Available')}</InlineNotification>
+            ) : (
+              <FormGroup>
+                <Dropdown
+                  titleText={t('selectForm', 'Select Form')}
+                  id="form-component-dropdown"
+                  label={t('selectForm', 'Select Form')}
```
NethmiRodrigo, line 147:
```suggestion
                  label={t('selectForm', 'Select form')}
```

src/components/interactive-builder/modals/add-form-reference/add-form-reference.modal.tsx
```diff
@@ -0,0 +1,241 @@
...
+              <InlineLoading description={t('loading', 'Loading...')} />
+            ) : clobdataError ? (
+              <InlineNotification>{t('errorLoadingForm', 'Error loading Form')}</InlineNotification>
+            ) : pages && pages.length > 0 ? (
+              <>
+                <FormGroup>
+                  <Dropdown
+                    titleText={`${selectedForm.name} ${t('pages', 'Pages')}:`}
+                    id="form-Pages-dropdown"
+                    label={t('Page', 'Select Page')}
```
NethmiRodrigo, line 168:
```suggestion
                    label={t('Page', 'Select page')}
```

src/components/interactive-builder/modals/add-form-reference/add-form-reference.modal.tsx
```diff
@@ -0,0 +1,241 @@
...
+                        />
+                      ))}
+                    </RadioButtonGroup>
+                  </FormGroup>
+                ) : null}
+              </>
+            ) : null}
+            {selectedSection ? (
+              <FormGroup>
+                <CheckboxGroup legendText={t('selectQuestions', 'Select Questions')}>
```
NethmiRodrigo, line 199:
```suggestion
                <CheckboxGroup legendText={t('selectQuestions', 'Select questions')}>
```

NethmiRodrigo (commented):
Thanks @MohamedAliMefteh, looks good overall! I haven't tested out the code, but here's a few comments from when I skimmed over this, I also noticed that a few translation keys haven't been added to the translation file, so please double check that too:

MohamedAliMefteh:
> Thanks @MohamedAliMefteh, looks good overall! I haven't tested out the code, but here's a few comments from when I skimmed over this, I also noticed that a few translation keys haven't been added to the translation file, so please double check that too:

@NethmiRodrigo, I think this pr still missing some few things and I am not sure should I keep the add reference feature to use normal forms or should it use only form components or both?

NethmiRodrigo:
> I think this pr still missing some few things 

Could you mention what those are?

> I am not sure should I keep the add reference feature to use normal forms or should it use only form components or both?

@samuelmale can we reference any form in another form, or does it have to be a component. According to what has currently being done on this PR, the form engine can render when the section of a normal form is referenced. 
For context, when I say normal form, I mean forms that haven't been made to be reusable as per the wiki - https://openmrs.atlassian.net/wiki/spaces/docs/pages/68780087/Core+Concepts#Creating-Components, i.e, doesn't have a `component_` in the form name, or has the encounter set to `component`.

MohamedAliMefteh:
> > I think this pr still missing some few things
> 
> Could you mention what those are?

The saving option as a component within the form builder, as well as the component filter in the forms dashboard, are still missing. However, these features should only be implemented if we decide to proceed with using form components. If we continue referencing standard forms instead, there is no need to include these two options.

MohamedAliMefteh:
The encounter types fetched from [this hook](https://github.com/openmrs/openmrs-esm-form-builder/blob/82636fb99a8f468eaa0fcecded5772eaec66d491/src/components/interactive-builder/modals/save-form/save-form.modal.tsx#L48C34-L49C3) in the Save Form modal do not include component as an option.
`const { encounterTypes } = useEncounterTypes();`

NethmiRodrigo:
> @samuelmale can we reference any form in another form, or does it have to be a component? According to what has currently being done on this PR, the form engine can render when the section of a normal form is referenced. For context, when I say normal form, I mean forms that haven't been made to be reusable as per the wiki - https://openmrs.atlassian.net/wiki/spaces/docs/pages/68780087/Core+Concepts#Creating-Components, i.e, doesn't have a `component_` in the form name, or has the encounter set to `component`.

Also looping in @ibacher

ibacher:
So, long-ish commentary here:

We have two mechanisms for loading content from different forms into an existing form, largely because one feature was developed without consulting the other: subforms and form references. Subforms (at least last I looked) are what the form engine itself supports. Form references are the version from the Angular version of the form engine.

Subforms apply only to pages and basically import the full page, including any form processors, from a different form.

Form references can be made to pages, sections, or individual questions, and will load the relevant page, section, or individual question from the referenced form. These are _not_ supported by the React form engine.

In neither case is there any restriction on how the form is named or the encounter type or anything else; those are just conventions.

All that's to say, I don't think the form engine itself supports this syntax, so I don't see a lot of value in adding support to it for the form builder.

MohamedAliMefteh:
> So, long-ish commentary here:
> 
> We have two mechanisms for loading content from different forms into an existing form, largely because one feature was developed without consulting the other: subforms and form references. Subforms (at least last I looked) are what the form engine itself supports. Form references are the version from the Angular version of the form engine.
> 
> Subforms apply only to pages and basically import the full page, including any form processors, from a different form.
> 
> Form references can be made to pages, sections, or individual questions, and will load the relevant page, section, or individual question from the referenced form. These are _not_ supported by the React form engine.
> 
> In neither case is there any restriction on how the form is named or the encounter type or anything else; those are just conventions.
> 
> All that's to say, I don't think the form engine itself supports this syntax, so I don't see a lot of value in adding support to it for the form builder.

Thanks for the explanation @ibacher . I was wondering—in this PR, i am using from reference as mentioned in [the docs](https://openmrs.atlassian.net/wiki/spaces/docs/pages/68780087/Core+Concepts#Creating-Components). It looks like the form engine was able to reference those forms, and I noticed that form reference seems to apply only to sections.

samuelmale:
>These are not supported by the React form engine.

@ibacher IIRC, RFE has native support for reference components.

ibacher:
> @ibacher IIRC, RFE has native support for reference components.

Yeah, I searched for the wrong thing. It does, in fact, have support for them!

ibacher:
> I noticed that form reference seems to apply only to sections

That's just what the examples show. I don't think it's a limitation of the functionality.

MohamedAliMefteh:
I have a unit test file for this feature, should I push it here or just create another PR.
@NethmiRodrigo

NethmiRodrigo:
> I have a test file for this feature, should I push it here or just create another PR. @NethmiRodrigo

You should do it in this one.

MohamedAliMefteh:
> You should do it in this one. Please add some e2e tests too.

I have noticed there is already a create-form-with-interactive-builder test file, and this feature is part of the interactive builder so I should add my e2e tests to this file, right?

NethmiRodrigo:
> I have noticed there is already a create-form-with-interactive-builder test file, and this feature is part of the interactive builder so I should add my e2e tests to this file, right?

Makes sense

src/components/interactive-builder/modals/add-form-reference/add-form-reference.modal.tsx
```diff
@@ -0,0 +1,243 @@
...
+  Checkbox,
+  RadioButtonGroup,
+  RadioButton,
+  Button,
+} from '@carbon/react';
+import { type Form as FormType, type Schema, type Page, type Section, type Question } from '@types';
+import { useForms } from '@hooks/useForms';
+import { useClobdata } from '@hooks/useClobdata';
+import { showSnackbar } from '@openmrs/esm-framework';
+import styles from './add-form-reference.scss';
```
NethmiRodrigo, line 23: Could you please organize these imports:

> The recommended order is:
> 
> React and framework imports (e.g., React, useState, useEffect)
> External modules (e.g., lodash, dayjs, react-i8next)
> Carbon component imports (e.g., Button, InlineLoading)
> OpenMRS imports (e.g., @openmrs/esm-framework, @openmrs/esm-patient-common-lib)
> Local imports (components, hooks, utilities, etc.)
> Asset imports (e.g import styles from './user.scss')

NethmiRodrigo (commented):
One more thing, we should probably remove the `referencedForms` entry if we delete the referenced section in the form.

MohamedAliMefteh:
> One more thing, we should probably remove the `referencedForms` entry if we delete the referenced section in the form.

yeah, absolutely, I forgot about it.

NethmiRodrigo (approved):
Hey @MohamedAliMefteh! This is really great work so far, and looks good to me, but would be great if we for referenced components, the edit button would reuse the `add-form-reference-modal` to be able to edit this, so a user would be able to update say the excluded questions.  
Screenshot 2025-05-14 at 8 27 49 PM

Since this is sort of big PR already, we can merge this in and you can work on this feature on a new PR

MohamedAliMefteh:
> Since this is sort of big PR already, we can merge this in and you can work on this feature on a new PR, or if you prefer we could do it in this one itself.

Let's keep it in this one since it makes sense for the functionality.
Now the question is—should we let the user update just the excluded questions, or allow them to edit the entire section(change the reference to another form for example)?

Either way, we could create an Edit Reference Modal, similar to the Edit Section Modal.
Alternatively, we could reuse the Add Reference Modal by introducing a `mode="edit"` prop. That way, it behaves differently when editing, and it would work well if we're only allowing updates to the excluded questions since in edit mode, we will disable the whole modal except for questions checkboxes.

NethmiRodrigo:
> should we let the user update just the excluded questions, or allow them to edit the entire section(change the reference to another form for example)?

I'd say edit all.

> we could reuse the Add Reference Modal by introducing a `mode="edit"` prop. 

I'd prefer this.

src/components/interactive-builder/modals/add-form-reference/add-form-reference.test.tsx
```diff
@@ -0,0 +1,345 @@
...
+  uuid: '',
+  version: '1.0',
+  pages: [
+    {
+      label: 'Test Page',
+      sections: [],
+    },
+  ],
+};
+describe('AddFormReferenceModal', () => {
```
NethmiRodrigo, line 127:
```suggestion
};

describe('AddFormReferenceModal', () => {
```

e2e/specs/create-form-with-interactive-builder.spec.ts
```diff
@@ -196,6 +203,34 @@ test('Create a form using the interactive builder', async ({ page, context }) =>
...
+      .getByRole('group', { name: /Sections:$/i })
+      .locator('span')
+      .nth(0)
+      .click();
+  });
+
+  await test.step('And then I click on `add`', async () => {
+    await expect(formBuilderPage.addButton()).toBeEnabled();
+    await formBuilderPage.addButton().click();
+    await expect(formBuilderPage.page.getByText(/Component added/i)).toBeVisible();
```
NethmiRodrigo, line 231: We should also check here that the JSON schema has the properties required, like how we do here - https://github.com/openmrs/openmrs-esm-form-builder/blob/a418ab95bbb99ea6645e37e42d86955cea35c67e/e2e/specs/create-form-with-interactive-builder.spec.ts#L99

NethmiRodrigo (commented):
Just some final minor nits, and then we're good to go

e2e/specs/create-form-with-interactive-builder.spec.ts
MohamedAliMefteh: > We should also check here that the JSON schema has the properties required, like how we do here -
> 
> https://github.com/openmrs/openmrs-esm-form-builder/blob/a418ab95bbb99ea6645e37e42d86955cea35c67e/e2e/specs/create-form-with-interactive-builder.spec.ts#L99

I am trying to access the content of the schema editor but i am having an error 
`SyntaxError: Expected ',' or '}' after property value in JSON at position 1177 (line 1 column 1178)`

this error is caused by the fact that the schemaText is not all the schema, it takes only the visible line of the schema,
I tried few solutions but none work i dont know if there is a way to handle

NethmiRodrigo: Yeah that's because playwright can only get what is within screen view and cannot scroll down to get the rest. The workaround for this, is to click the `Copy to clipboard` and get the JSON schema from there. See here for reference - https://github.com/openmrs/openmrs-esm-form-builder/blob/a418ab95bbb99ea6645e37e42d86955cea35c67e/e2e/specs/create-form-with-interactive-builder.spec.ts#L192-L196

NethmiRodrigo (approved):
Alrighty, thanks @MohamedAliMefteh!

PR #519, opened 2025-05-15
(feat) O3-3958: Add a select all option for concept answers

NethmiRodrigo:
## Requirements

## Summary

This PR adds the option `Select all` for when users are selecting answers from a concept.

## Screenshots

https://github.com/user-attachments/assets/38ad795a-6d13-4fac-ace5-a48e291e3bd0

## Related Issue

https://issues.openmrs.org/browse/O3-3958

### DIFF
--- a/src/components/interactive-builder/modals/question/question-form/rendering-types/inputs/select/select-answers.component.tsx
+++ b/src/components/interactive-builder/modals/question/question-form/rendering-types/inputs/select/select-answers.component.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useMemo, useState } from 'react';
+import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { InlineNotification, MultiSelect, Stack } from '@carbon/react';
 import { DndContext, type DragEndEvent } from '@dnd-kit/core';
 import { arrayMove, SortableContext, verticalListSortingStrategy } from '@dnd-kit/sortable';
@@ -20,6 +20,7 @@ const SelectAnswers: React.FC = () => {
   const { formField, concept, setFormField } = useFormField();
   const [addedAnswers, setAddedAnswers] = useState<AnswerItem[]>([]);
   const [invalidAnswerIds, setInvalidAnswerIds] = useState<string[]>([]);
+  const initializedConceptUuid = useRef<string | null>(null);
 
   // Reset the additionally selected answers when the concept changes
   useEffect(() => {
@@ -28,9 +29,16 @@ const SelectAnswers: React.FC = () => {
     }
   }, [concept]);
 
-  // Initialize the selected answers to be the concept's answers
+  // Initialize the selected answers to the concept's answers, once per concept.
+  // The once-per-concept guard keeps a cleared selection cleared instead of
+  // re-selecting every answer whenever the answers list becomes empty.
   useEffect(() => {
-    if (concept?.answers?.length && !formField.questionOptions?.answers?.length) {
+    if (!concept || initializedConceptUuid.current === concept.uuid) {
+      return;
+    }
+    initializedConceptUuid.current = concept.uuid;
+
+    if (concept.answers?.length && !formField.questionOptions?.answers?.length) {
       const initialAnswers = concept.answers.map((answer) => ({
         concept: answer.uuid,
         label: answer.display,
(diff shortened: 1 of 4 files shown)


ibacher (approved):

ibacher:
Code here is fine, but UX question: do you think it maybe makes sense for all possible Concept Answers to be selected by default so that people are basically de-selecting answers they don't want?

NethmiRodrigo:
> Code here is good, but UX question: do you think it maybe makes sense for all possible Concept Answers to be selected by default so that people are basically de-selecting answers they don't want?

Makes sense to me!

This made me also think, the reason for having a "Select All" option was for cases where there would be a lot of answers, would it also be good to have this answer dropdown be searchable?

ibacher:
> would it also be good to have this answer dropdown be searchable?

Definitely! Especially because people have a bad habit of doing something like adding all ICD-11 codes as an answer...

NethmiRodrigo:
> Definitely! Especially because people have a bad habit of doing something like adding all ICD-11 codes as an answer...

Turns out the `FilterableMultiSelect` component doesn't support having `Select All` out of the box...

src/components/interactive-builder/modals/question/question-form/rendering-types/inputs/select/select-answers.component.tsx
```diff
@@ -19,12 +19,31 @@ const SelectAnswers: React.FC = () => {
...
+
+      setFormField((prevField) => ({
+        ...prevField,
+        questionOptions: {
+          ...prevField.questionOptions,
+          answers: initialAnswers,
+        },
+      }));
+    }
+  }, [concept, formField.questionOptions?.answers, setFormField]);
```
denniskigen, line 55: Nice work implementing the select-all-by-default suggestion, and having checked the form engine, writing the concept answers into the schema is exactly right since radio/dropdown/multi-select have no runtime fallback to concept answers. One behavior worth fixing before merge: because this effect re-runs whenever the answers list becomes empty (the deps include `formField.questionOptions?.answers`), clearing the selection instantly re-selects everything. The MultiSelect exposes a clear-selection control, so an author who wants a small subset of a large answer list, like the ICD-11 case discussed above, can never clear and start from zero. Could we initialize only when the concept first arrives (for example, track the last-initialized concept uuid in a ref) instead of whenever answers are empty? Guarding against saving with zero answers feels like a separate validation concern rather than something the initializer should enforce.

src/components/interactive-builder/modals/question/question-form/rendering-types/inputs/select/select-answers.component.tsx
```diff
@@ -170,7 +202,7 @@ const SelectAnswers: React.FC = () => {
 
   return (
     <Stack gap={5}>
-      {answerItems.length > 0 && (
+      {answerItems.length > 1 && (
```
denniskigen, line 269: This guard makes sense for hiding a dropdown that contains only the select-all item. One edge it introduces: the early-return branch above (concept with no answers plus saved custom answers) does not append the select-all item, so a question with exactly one saved custom answer no longer renders the MultiSelect and that answer cannot be deselected on re-edit. Appending the select-all item in that branch too would fix it, and Carbon sorts `isSelectAll` items to the top regardless of where they appear in the array.

src/components/interactive-builder/modals/question/question-form/rendering-types/inputs/select/select-answers.component.tsx
```diff
@@ -131,7 +152,15 @@ const SelectAnswers: React.FC = () => {
         text: answer.label,
       }));
 
-    return [...answersFromConceptWithLabelsFromFormField, ...additionalAnswers];
+    return [
+      ...answersFromConceptWithLabelsFromFormField,
+      ...additionalAnswers,
+      {
+        id: 'select-all',
+        text: 'Select all options',
```
denniskigen: Tiny one: `'Select all options'` should go through translation like the rest of this component, e.g. `t('selectAllOptions', 'Select all options')` with a matching `en.json` entry.

denniskigen (commented):
Took a fresh look at this since it has been open for a while. The approach holds up well: I checked the form engine and `questionOptions.answers` is the sole runtime source of options for radio/dropdown/multi-select, with no fallback to concept answers, so initializing the schema with the concept answers is the right mechanism. Three inline comments below, one behavioral and two small. Also, a rebase is needed before this can land: the component was touched by #403 and #1086, and the test file has since migrated to vitest (#1198), so the updated test needs porting.

src/components/interactive-builder/modals/question/question-form/rendering-types/inputs/select/select-answers.component.tsx
denniskigen: Addressed in 1139c67. The initializer now runs once per concept, tracked via a ref, so a cleared selection stays cleared. Added a regression test that fails against the previous behavior.

src/components/interactive-builder/modals/question/question-form/rendering-types/inputs/select/select-answers.component.tsx
denniskigen: Addressed in 1139c67. The custom-answers-only branch now appends the select-all option as well, so the menu still renders for a single saved custom answer. Added a regression test.

src/components/interactive-builder/modals/question/question-form/rendering-types/inputs/select/select-answers.component.tsx
denniskigen: Addressed in 1139c67 with a new selectAllOptions translation key.

PR #552, opened 2025-06-06
(chore) O3-4788: Standardize config schema and improve type safety

harshthakkr:
## Requirements

## Summary
This PR improves the config schema file in the `esm-form-builder` repository by:

- Adding proper types and default values
- Defining appropriate validators where needed
- Updating the corresponding TypeScript interfaces to reflect these changes

## Related Issue
[O3-4788](https://openmrs.atlassian.net/browse/O3-4788)

[O3-4788]: https://openmrs.atlassian.net/browse/O3-4788?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/config-schema.ts
```diff
@@ -43,18 +46,29 @@ export const configSchema = {
...
+    _elements: {
+      _type: Type.String,
+    },
   },
   showSchemaSaveWarning: {
     _type: Type.Boolean,
     _default: true,
     _description: 'Whether to show a warning about possibly losing data in the forms dashboard',
   },
   dataTypeToRenderingMap: {
```
denniskigen, line 58: Can we restore the `_description` and `_default` properties here?

harshthakkr: 1. @vasharma05 suggested that I place `_default` as deep as possible within the object structure.
2. For some reason, including only `_description` causes the entire app to crash when I hover over the property.

https://github.com/user-attachments/assets/90bf07e2-c797-45ae-9b12-709e35385bf8

src/config-schema.ts
```diff
@@ -46,12 +52,11 @@ export const configSchema = {
   },
   showSchemaSaveWarning: {
     _type: Type.Boolean,
-    _default: true,
     _description: 'Whether to show a warning about possibly losing data in the forms dashboard',
+    _default: true,
   },
   dataTypeToRenderingMap: {
     _description: 'A map used to match concept datatypes to rendering types',
-    _type: Type.Object,
```
denniskigen, line 54: Why are we getting rid of the type here?

src/config-schema.ts
```diff
@@ -79,16 +84,70 @@ export const configSchema = {
...
+      _type: Type.Array,
+      _elements: {
+        _type: Type.String,
+      },
+    },
   },
   enableFormValidation: {
     _type: Type.Boolean,
-    _default: false,
     _description: 'Whether to enable form validation',
```
denniskigen, line 144:
```suggestion
    _description: 'Whether to block form rendering when there are validation errors',
```

src/config-schema.ts
```diff
@@ -46,12 +52,11 @@ export const configSchema = {
...
   showSchemaSaveWarning: {
     _type: Type.Boolean,
-    _default: true,
     _description: 'Whether to show a warning about possibly losing data in the forms dashboard',
+    _default: true,
   },
   dataTypeToRenderingMap: {
     _description: 'A map used to match concept datatypes to rendering types',
-    _type: Type.Object,
     _default: {
```
denniskigen, line 60: It looks like we're mixing `_default` values with individual property schemas at the same level. I think we should either go with the simpler approach:

```ts
  dataTypeToRenderingMap: {
    _type: Type.Object,
    _description: 'A map used to match concept datatypes to rendering types',
    _default: { /* existing defaults */ },
  },
```

or the more complex approach:

```ts
  dataTypeToRenderingMap: {
    _type: Type.Object,
    _description: 'A map used to match concept datatypes to rendering types',
    _default: { /* your existing defaults */ },
    _elements: {
      _type: Type.Array,
      _elements: { _type: Type.String }
    }
  },
```

but not a mix of both. Maybe the simpler approach works better here?

harshthakkr: FYI `_elements` is only allowed for `_type: Type.Array`

src/config-schema.ts
brandones: I'm guessing this comment is outdated?

src/config-schema.ts
brandones: OH WAIT. **Ignore all that**. I just realized this isn't supposed to be a `Type.Object` at all. Everything is fine. Let's still add the array element validators though.

ibacher: Technically, `ConceptDatatype` is something implementations can extend, but I've never seen that in any real-world implementation. If anything, Concept datatypes are ignored (e.g., N/A concepts used to store values) or overloaded.

brandones: Is there any way that they could provide a custom handler given the way things are coded right now? If not, I think we should just hard-code the concept datatypes until we decide to actually properly support custom datatypes.

brandones: Ok my inclination is to switch off of `Type.Object` and to using just a normal config tree. Does that make sense, @harshthakkr ?

src/config-schema.ts
```diff
@@ -1,60 +1,88 @@
...
-      'encounterProvider',
-      'encounterRole',
-      'obs',
-      'obsGroup',
-      'patientIdentifier',
-      'personAttribute',
-      'testOrder',
-      'programState',
-    ],
+    _default: allowedQuestionTypes,
```
brandones, line 54: I am unsure how the authors of esm-form-builder intended for this to be used. But you don't need to worry about that :)

src/config-schema.ts
brandones: Yeah now that I'm reading this config schema as you've rewritten it, it is definitely a shame that the upper-level descriptions aren't working, because it would indeed be helpful here.

brandones (approved):
Great work @harshthakkr ! This is a big improvement all around.

src/config-schema.ts
harshthakkr: Yup, I have made changes assuming it's a config tree element.

PR #725, opened 2025-07-23
(feat) O3-4899- Implement backend translation fetching and merging in Translation Builder

Bharath-K-Shetty:
## Requirements

## Summary

Overview:
This PR implements the feature to fetch saved translation files from the backend for a selected language and intelligently merge them with the current form schema in the Translation Builder. It ensures that only valid translations (i.e., keys still present in the schema) are retained and used.

This PR covers:

- Fetching translation files from the backend for the selected language.

- Merging backend translations with the current schema using extractTranslationStringsFromSchema.

- Only including translations whose keys are still present in the schema.

- Ignoring outdated keys and using schema strings as fallback for missing translations.

## Related Issue

https://openmrs.atlassian.net/browse/O3-4899

### DIFF
--- a/src/components/translation-builder/translation-builder.component.tsx
+++ b/src/components/translation-builder/translation-builder.component.tsx
@@ -4,6 +4,8 @@ import { InlineLoading, InlineNotification, IconButton, Tabs, Tab, TabList, Drop
 import { Download, Edit, ArrowRight } from '@carbon/react/icons';
 import { showModal } from '@openmrs/esm-framework';
 import { useLanguageOptions } from '@hooks/getLanguageOptionsFromSession';
+import { useBackendTranslations } from '@hooks/useBackendTranslations';
+
 import { extractTranslatableStrings } from '../../utils/translationSchemaUtils';
 import styles from './translation-builder.module.scss';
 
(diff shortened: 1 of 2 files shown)


src/components/translation-builder/translation-builder.component.tsx
```diff
@@ -109,6 +104,39 @@ const TranslationBuilder: React.FC<TranslationBuilderProps> = ({ formSchema, onU
     },
     [fallbackStrings],
   );
+  useEffect(() => {
+    if (selectedLanguageCode === 'en' && formSchema) {
+      setTranslations(fallbackStrings);
+    }
+  }, [selectedLanguageCode, formSchema, fallbackStrings]);
```
NethmiRodrigo, line 111: If this is what needs to be done if the language is english, then why can't it be inside the `languageChanger` function itself. I don't think this requires a `useEffect`. If you need to this to be the case when the app mounts you could instantiate the `translations` string with `fallbackStrings`

NethmiRodrigo (commented):
Good work @Bharath-K-Shetty! A minor thing:

src/components/translation-builder/translation-builder.component.tsx
Bharath-K-Shetty: It can’t be inside `languageChanger` because that function only runs when the language is explicitly changed. On initial mount, it doesn’t get triggered unless we call it manually. Also, I’m not initializing the translations state with fallbackStrings because formSchema is loaded asynchronously -  so at the time of initialization, it’s not available, and fallbackStrings can't be generated yet. That's why I’m using` useEffect` to ensure it updates correctly once formSchema is available.

src/components/translation-builder/translation-builder.component.tsx
```diff
@@ -109,6 +104,39 @@ const TranslationBuilder: React.FC<TranslationBuilderProps> = ({ formSchema, onU
     },
     [fallbackStrings],
   );
+  useEffect(() => {
```
NethmiRodrigo, line 107:
```suggestion

  useEffect(() => {
```

src/components/translation-builder/translation-builder.component.tsx
```diff
@@ -109,6 +104,39 @@ const TranslationBuilder: React.FC<TranslationBuilderProps> = ({ formSchema, onU
...
+    if (selectedLanguageCode === 'en' && formSchema) {
+      setTranslations(fallbackStrings);
+    }
+  }, [selectedLanguageCode, formSchema, fallbackStrings]);
+
+  const languageChanger = async (newLangCode: string) => {
+    setSelectedLanguageCode(newLangCode);
```
NethmiRodrigo, line 114: Shouldn't the loading be set to true at the very start?
```suggestion
    setLoading(true);
    setSelectedLanguageCode(newLangCode);
```

Bharath-K-Shetty: No, @NethmiRodrigo, we shouldn't do that because if the selected language is English, the loading state will remain true indefinitely, causing an infinite loading issue.

NethmiRodrigo (approved):

NethmiRodrigo:
Thanks @Bharath-K-Shetty!

PR #1335, opened 2026-09-14
(feat) O3-4997: Differentiate retired concepts in the form builder

VeronicaMuthee:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-docs.openmrs.org/en-US/docs/frontend-modules/contributing/#contributing-guidelines) label. See existing PR titles for inspiration.
- [x] My work is based on designs, which are linked or shown either in the Jira ticket or the description below. (See also: [Styleguide](http://om.rs/o3ui))

## Summary
Retired concepts now appear in concept search results (via includeAll=true) rendered with dimmed typography and a "Retired" tag, while remaining selectable so legacy forms can still be edited.
- Problem: the form builder gave no indication a concept was retired — and the REST search hides retired concepts by default, so a maintainer editing a legacy form couldn't even find its backing concept.
- Changes: includeAll=true in useConceptLookup; dimmed text + cool-gray "Retired" tag in the shared ConceptSearch dropdown (still clickable, per the ticket's guidance); dimmed input text when the selected/pre-loaded backing concept is retired; retired?: boolean on the Concept type. Form preview and form engine untouched.

## Screenshots
image

https://github.com/user-attachments/assets/c2491523-625a-4694-a144-bc2a1f8dabda

## Related Issue

- https://openmrs.atlassian.net/browse/O3-4997


src/components/interactive-builder/modals/question/question-form/common/concept-search/concept-search.component.tsx
```diff
@@ -135,11 +140,18 @@ const ConceptSearch: React.FC<ConceptSearchProps> = ({
                 {concepts?.map((concept, index) => (
                   <li
                     role="menuitem"
-                    className={styles.concept}
+                    className={classNames(styles.concept, {
+                      [styles.retiredConcept]: concept.retired,
+                    })}
```
ibacher: This is an irrelevant nit, but this might be more cleanly expressed using a data attribute, e.g.,

```jsx
<li data-retired={concept.retired}>
```

Then the styling becomes:

```scss
.concept {
  // as before

  &[data-retired] {
    color: colors.$gray-50;
  }
}
```

src/components/interactive-builder/modals/question/question-form/common/concept-search/concept-search.scss
```diff
@@ -30,6 +33,15 @@
   }
 }
 
+// Retired concepts render as if disabled typographically, but remain clickable
+.retiredConcept {
+  color: colors.$gray-50;
```
ibacher: nit: I think the right token here is `$text-disabled` from `@carbon/themes`.

NethmiRodrigo (approved):
Thanks @VeronicaMuthee! LGTM, I made a minor change to how the retired concept shows up. In the dropdown, it no longer appears grayed out and the tag is right next to the text (rather than at the corner of the list item), and if a retired concept is selected, it now shows `(Retired)` next to the text. Let me know if you don't like how it looks :D

https://github.com/user-attachments/assets/45c51611-119e-47f5-a2b8-e2e37fd8a371

src/components/interactive-builder/modals/question/question-form/common/concept-search/concept-search.scss
NethmiRodrigo: Aah sorry, I didn't see the review you left and changed it up to remove the disabled styling because it gave the impression that the menu item couldn't be selected...

ibacher: I'm ok with giving the impression that it couldn't be selected. It shouldn't be selected.

ibacher (approved):

